PR #136, opened 2014-02-14
Adds a third option for strict to raise a Validation Error

wolfe:
I supplied this third option so that Django Rest Framework could return a 400 w/ an appropriate error message.

If this is to your liking, Alex, I'd be glad to add some unit tests before you merge.

Note the bizarre constants True, False, "RAISE" -- I did this to support backward compatibility.  If you recommend a cleaner way to manage the strict options, let me know. Someone here recommended another boolean option, but then there'd be 4 possible values of the two booleans, only three of which make sense.  I think this is more robust.


mdentremont:
@carltongibson, I've taken @wolfe's code and reapplied it on top of your develop branch.

I also made some tweaks to get the tests passing.

Would you be able to take a peek and let me know if I missed anything?

Thanks!

carltongibson:
@mdentremont — That looks good. I've been totally snowed under with work but am planning on putting a release together over (say) the next week. I'll review properly and we'll get this in. 

Awesome. Thanks!

mdentremont:
@carltongibson Thanks for the quick reply, sounds great!

django_filters/filterset.py
```diff
@@ -266,7 +276,8 @@ def __new__(cls, name, bases, attrs):
 class BaseFilterSet(object):
     filter_overrides = {}
     order_by_field = ORDER_BY_FIELD
-    strict = True
+    # What to on on validation errors
```
carltongibson: Typo here.

carltongibson:
@wolfe @mdentremont Guys this is great. 

I just have one issue remaining: can we add just a little to the usage docs explaining the new option? (It needn't be long but the docs are my main concern and I don't want them to get _less comprehensive_).

mdentremont:
@carltongibson Any suggestions on where the docs should go?

mdentremont:
@carltongibson I have addressed the typo and added the usage docs!

wolfe:
Thank you, Matt, for taking this on!

carltongibson:
Good work both of you! I'll review and merge. Thanks!

carltongibson:
@wolfe, @mdentremont — Thanks for this gentlemen. Very good.

adam-snyder:
Thank you for adding this feature. How soon can we expect it in the next release? Keep up the good work.

carltongibson:
I want to roll in #264 and then I'll look at a new release. (Until then you can use `develop`)

adam-snyder:
Sounds good. I have to wait for it to be published to pypi before I can use it for our deployment. Thanks again.

carltongibson:
@adam-snyder You know you can `pip install ...` straight from a GitHub branch right? https://pip.pypa.io/en/latest/reference/pip_install.html#git

restless:
What's the purpose of displaying _("Filter") as the default help text? I've just upgraded to latest django-filter and suddenly all my filter fields started to display extra "Filter" text. I found that this comes from this pull request (https://github.com/aioTV/django-filter/commit/90d244b8277f7b22e7761e9913f259c11a7a91b8).

wolfe:
I don't recall the specifics... 'twas a while ago.  But I agree it should
have been a separate pull request.  Of course, the main point of the pull
request was to pay attention to the argument help_tex, so you can set
help_text='' if you prefer.

I do prefer consistency --- if help_text is auto-injected for one kind of
filter (the exclusion filter) it should be for all. I vaguely recalled that
this resulted in the best experience when combined with Django REST
Framework which can potentially auto-generate lots of fields with
auto-generated help text.

But I appreciate that the change was backward incompatible, and so may have
been ill-advised.  Perhaps settings like DJANGO_FILTER_FILTER_HELP_TEXT and
DJANGO_FILTER_EXCLUSION_HELP_TEXT are advised?

On Mon, Aug 24, 2015 at 11:24 PM, Jakub Wiśniowski <notifications@github.com

restless:
Thanks for explanation wolfe. Probably it's more reasonable in case of DRF than in case of a regular filter. I'd say that:

1. I think that no value for a regular filter field is ok and doesn't break consistency. I'd compare it to a situation where a developer is encouraged to avoid commenting the obvious code. The exclusion filter is breaking the obviousness about how filter works so it has its "comment".

2. Extra settings would work for me too

My preference is to go with 1 unless you think that it's good to keep the "filter' for DRF. It's just my 2 cents, as I think that some action is required here to avoid forcing users to override all help_texts in all their filters.

ad-m:
:-1: for any change here. I want avoid write everywhere ```help_text=""```. I was also surprised by the change.

carltongibson:
OK. We've got competing use-cases here. I'll add a flag to FilterSet (and propagate that to Filter) to enable/disable auto-populating the help text. This will keep it nice and declarative. 

I'll have the default be disable auto-populating — I think that's less surprising.

Unless someone jumps in with a PR, that'll not be instantly. 

In the meantime, rather than downgrade, I suggest overriding `__init__` on your `FilterSet` to loop through `filters` and set the `extras["help_text"]` key there.

rjschave:
I too was surprised by this change.  I ended up doing what Carlton suggested with something like this:

```python
# specify empty string for help text to suppress on the filter form
for key in self.filters.iteritems():
    self.filters[key[0]].extra.update(
        {'help_text': ''}
    )
```

PR #248, opened 2015-05-20
More consistent handling of distinct=True and exclude=True across filters

JonPeel:
The parameters available in the base Filter class don't seem to be consistently adhered to - the intention here is to try and remedy some of that.

I've skipped DateRangeFilter as dealing with the class level lambda's is an interface change I don't want to get involved in.  Everything else is rather straightforward.

Rolls in fixes for merge conflicts to: https://github.com/alex/django-filter/pull/217


JonPeel:
Closing - created from wrong branch.

PR #388, opened 2016-03-14

byrman:


rpkilby:
Hi @byrman - it looks like you've forked off of the master branch, which is used for releases. You may want to fork off of develop instead.

tests/test_filtering.py
```diff
@@ -300,6 +302,102 @@ class Meta:
         self.assertQuerysetEqual(f.qs, [2], lambda o: o.pk)
 
 
+@unittest.skipIf(
+    django.VERSION < (1, 8), "version does not support DurationField")
```
rpkilby: btw, this check isn't really necessary since django 1.8 is the minimum supported version

byrman:
Thanks for your feedback. I'll remove the 2 lines, make a pull request based on develop, and close this one.

PR #390, opened 2016-03-14

byrman:


carltongibson:
@byrman This looks good. We just need an entry in the Filters reference docs... :-)

byrman:
Thanks. I'll will add an entry in the docs.

carltongibson:
@byrman sorry, I didn't mean to close the PR. I must have hit the wrong button.

byrman:
No problem. I opened a new one, so this one can remain closed.

PR #393, opened 2016-03-22
Add Meta.fields compatibility with annotated fields from the default query

rpkilby:
This is a little convoluted, but I think it would be a valuable addition. 

#### Context:
Field annotations give us the flexibility of computed properties on our models. eg, you could have a user stats model that aggregates the comment count per user. You could then filter the user stats for the more prolific commenters on a blog. 

While these annotations can be filtered on, they are not proper model fields and cannot be resolved during `Meta.fields` processing. However, there is effectively a 'default queryset' which can be used to retrieve these annotations.

#### Proposal:
Allow `Meta.fields` to parse annotations under the following conditions:
- the annotation is performed in the default manager's `get_queryset()` method.
- the manager must be enabled for related fields with `use_for_related_fields`.

**Update:**
Apparently filtering annotations across relationships is not currently supported. Filed a bug [here](https://code.djangoproject.com/ticket/26393). The updated `Meta.fields` handling can correctly generate the related filter class, but its unusable.

TODO:
- [x] Add failing usecase.
- [x] implement
- [x] test
- [ ] docs
- [ ] wait for feedback on [26393](https://code.djangoproject.com/ticket/26393).

### DIFF
--- a/django_filters/utils.py
+++ b/django_filters/utils.py
@@ -60,7 +60,16 @@ def get_model_field(model, field_name):
     try:
         return opts.get_field(parts[-1])
     except FieldDoesNotExist:
+        pass
+
+    annotations = get_model_annotations(opts.model)
+    if annotations is None or parts[-1] not in annotations:
         return None
+
+    # get output_field and bind model so it behaves like a proper model field.
+    field = annotations[parts[-1]].output_field
+    field.model = opts.model
+    return field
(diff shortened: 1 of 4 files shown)


rpkilby:
Closing this since you cannot filter annotations across relationships. Completely misunderstood how that worked.

PR #407, opened 2016-04-16
Lookup types

mjschultz:
It seems as though Django 1.9 has become more particular about query lookups because I was trying to implement a FilterSet such that it would enable all the lookup types for the field. Looking through the issue/pr tracker [using the `LOOKUP_TYPES` variable](https://github.com/carltongibson/django-filter/pull/301) was suggested. I tried to do that and got some errors.

I ended up adding a test that implements the example code @carltongibson has in that PR to verify that it wasn't caused by some crazy DRF wrapping (it wasn't). Then I figured I'd make the test pass.

I don't know if my fix is the best, but I think the test should be okay and it should pass in any case.

### DIFF
--- a/django_filters/filterset.py
+++ b/django_filters/filterset.py
@@ -8,6 +8,7 @@ from collections import OrderedDict
 from django import forms
 from django.forms.forms import NON_FIELD_ERRORS
 from django.core.validators import EMPTY_VALUES
+from django.core.exceptions import FieldError
 from django.db import models
 from django.db.models.constants import LOOKUP_SEP
 from django.db.models.fields.related import ForeignObjectRel
@@ -426,7 +427,10 @@ class BaseFilterSet(object):
 
     @classmethod
     def filter_for_field(cls, f, name, lookup_expr='exact'):
-        f, lookup_type = resolve_field(f, lookup_expr)
+        try:
+            f, lookup_type = resolve_field(f, lookup_expr)
+        except FieldError:
+            return
 
         default = {
             'name': name,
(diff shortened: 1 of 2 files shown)


mjschultz:
I should have seen those failures coming based on my description of django 1.9 being more strict about lookup types.

As such, the test now allows django < 1.9 to have more `base_filters` while 1.9+ has fewer.

rpkilby:
Hi @mjschultz. If you're interested in exposing all lookups, I recommend looking at [drf-filters](https://github.com/philipn/django-rest-framework-filters). There is an `AllLookupsFilter` and an `'__all__'` shorthand for the dict-style syntax that does this correctly.

Using `LOOKUP_TYPES` ends up actually have the wrong behavior. In django 1.8, the additional lookups are invalid. Django used to register lookups for all of the default query terms. This mostly worked, but allowed you to do nonsensical things like perform date/time lookups on an integer field. Django 1.9 has a smaller set of default, but universally applicable lookups. Specific lookups, such as the date & time lookups are only registered on the date & time fields. Using `LOOKUP_TYPES` sort of works here again in Django 1.9, however you're missing out on the new transforms. In 1.8, you could only perform exact queries on the date/time lookups. eg, `published_on__year=2000`. In 1.9, transforms allow you to do `published_on__year__gt=2000`. `LOOKUP_TYPES` is unaware of the transforms. It's also not aware of any custom transforms and lookups.

As to the PR, I don't think hiding FieldError is the right thing to do. The error let's users know at class load time that something has been configured incorrectly. eg, a misspelled `excat` lookup. That said, the error is missing useful field/lookup information, and could be enhanced. Hope that helps!

PR #417, opened 2016-05-11
adding the support for distinct to the range query

porschelau:
I found the bug and i have submitted an issue


joinworldtv:
Hi there,

DISTINCT doesn't work, because it doesn't achieve  qs = qs.distinct()   string:

`
class RangeFilter(Filter):
    field_class = RangeField

    def filter(self, qs, value):
        if value:
            if value.start is not None and value.stop is not None:
                lookup = '%s__range' % self.field_name
                return self.get_method(qs)(**{lookup: (value.start, value.stop)})    <-----
....
`

You have to change to this:

`
qs = self.get_method(qs)(**{lookup: (value.start, value.stop)}) 
`

Best,
Alex

PR #432, opened 2016-06-16
Fix on_delete deprecation warning

rpkilby:


rpkilby:
Whoops - that model should be deleted instead, as it's superseded by `ManagerGroup`.

PR #451, opened 2016-08-02
Resolve 450, add error message for unrecognized field types

rpkilby:
This resolves the issue discussed in #450, maintaining forwards and backwards compatibility. 

New behavior:
- `Meta.fields` accepts either the list/dict syntax, or the '__all__' special value
- `Meta.exclude` can be used in lieu of `Meta.fields` (similar to Django's API).
- Not specifying either will result in no automatic filter generation (just uses declared filters).

From the original issue, went ahead and added the assertion error for unrecognized model field types such as postgres's db fields, which don't have mapped `filter_class`es.


tests/test_filterset.py
```diff
@@ -454,42 +459,40 @@ class Meta:
 
         self.assertEqual(set(F.base_filters), set(['name', 'serves_pizza']))
 
-    def test_custom_field_ignored(self):
-        class F(FilterSet):
```
rpkilby: Test moved to deprecations as `test_fields_not_set_ignore_unknown`.

PR #453, opened 2016-08-03
Update Django and Python versions in testing matrix

edmorley:
* Stop testing on Python 3.2 (since it's [no longer supported](https://docs.python.org/devguide/#status-of-python-branches) as of 2016-02-20)
* Remove Django 1.7 remnants (these were left behind when Django 1.7 was removed from Travis testing)
* Add a Django 1.8+Python 3.5 combo to the Tox config (this was already being tested on Travis)
* Test on Django 1.10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ cache: pip
 
 python:
   - "2.7"
-  - "3.2"
   - "3.3"
   - "3.4"
   - "3.5"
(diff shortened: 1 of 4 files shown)


rpkilby:
Side note - Django 1.8 will officially drop support for python 3.2 at the end of 2016 (discussed [here](https://docs.djangoproject.com/en/dev/releases/1.8.8/)).

PR #463, opened 2016-08-19
Simplify multiple choice filter.__init__

blueyed:


ad-m:
LGTM :+1:

django_filters/filters.py
```diff
@@ -195,12 +195,8 @@ class MultipleChoiceFilter(Filter):
     always_filter = True
 
     def __init__(self, *args, **kwargs):
-        distinct = kwargs.get('distinct', True)
```
rpkilby: You could use setdefault instead.
```py
kwargs.setdefault('distinct', True)
```

blueyed: Good point!  Done in https://github.com/carltongibson/django-filter/pull/468.

PR #466, opened 2016-08-19
Cleanup MultipleChoiceFilter.filter

blueyed:
Also some minor optimization included there.

### DIFF
--- a/django_filters/filters.py
+++ b/django_filters/filters.py
@@ -217,26 +217,26 @@ class MultipleChoiceFilter(Filter):
         return False
 
     def filter(self, qs, value):
-        value = value or ()  # Make sure we have an iterable
+        if not value:
+            # Even though not a noop, no point filtering if empty.
+            return qs
 
         if self.is_noop(qs, value):
             return qs
 
-        # Even though not a noop, no point filtering if empty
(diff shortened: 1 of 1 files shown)


django_filters/filters.py
```diff
@@ -217,26 +217,26 @@ def is_noop(self, qs, value):
         return False
 
     def filter(self, qs, value):
-        value = value or ()  # Make sure we have an iterable
-
-        if self.is_noop(qs, value):
+        if not value:
```
rpkilby: This seems sensible to me, however the standard filter methods uses `if value in ([], (), {}, None, '')`. I've always wondered why this was the case instead of just `if not value`.

rpkilby: Duh - generally you need to perform an empty value check, because there are valid falsey values such as... `False` that you may want to filter on. In the case of the multiple choice filter, the incoming value should be an iterable anyway, so special handling shouldn't be necessary.

blueyed: @rpkilby 
So you think that this is fine like it is?

rpkilby: Yeah I think so. If it's not fine, it's an easy fix later on.

Issue #148, opened 2014-06-23
Aggregation: possible?


ncharlot:
Hi all,

I have a simple "stats" model with a date, a type (related model) and a number.
I would like to add a "group by" filter allowing to group results by date or by type and summing the number.

Is there a way to do that?

Thx!

PR #484, opened 2016-09-14
Add settings conf, add deprecation warnings

rpkilby:
Adds a new conf module, deprecates `HELP_TEXT_*` settings in preparation for #437.

### DIFF
--- a/docs/migration.txt
+++ b/docs/migration.txt
@@ -153,3 +153,13 @@ on the FilterSet:
 
 To fix this, simply remove the methods from your class. You can subclass
 ``OrderingFilter`` to migrate any custom logic.
+
+
+Deprecated ``FILTERS_HELP_TEXT_FILTER`` and ``FILTERS_HELP_TEXT_EXCLUDE``
+-------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


docs/migration.txt
```diff
@@ -153,3 +153,10 @@ on the FilterSet:
 
 To fix this, simply remove the methods from your class. You can subclass
 ``OrderingFilter`` to migrate any custom logic.
+
+
+Deprecated ``FILTERS_HELP_TEXT_FILTER`` and ``FILTERS_HELP_TEXT_EXCLUDE`` settings
+----------------------------------------------------------------------------------
+Details: https://github.com/carltongibson/django-filter/pull/484
```
ad-m: There is no details .What I missing? I use this settings in my project (as a author of patch which introduce them). See #136, #148, #292 for user feedback.

rpkilby: Hi. I've updated the PR with more details. This PR is prep for #437, which includes more descriptive filter labels. eg, "Published date is less than" and "Exclude author name starts with". The `help_text` would no longer be necessary.

PR #500, opened 2016-09-23
Fix rest framework template handling

rpkilby:
This supersedes and closes #498. 

I've inlined the templates as strings, since they are so small. This means that it's unnecessary to add `'django_filters.rest_framework'` to the `INSTALLED_APPS` setting.


rpkilby:
> Inlining is OK here, but it kind of assumes nobody will ever override the templates. I suspect that's a bad assumption.

Good point. They could override, but it would require that they subclass the backend. I've updated the PR to check for the path first, then fallback to the inline templates. This keeps packaging simple and doesn't require adding to  `INSTALLED_APPS`.

PR #516, opened 2016-10-10
FilterSetMetaclass: look for order_by_field in attrs

blueyed:
`filters[order_by_field]` might come from a parent class, and the
automatic ordering filter needs to take into account the current class
and its filters (for `order_by=True`).

If this is sensible there should be a test for it probably.

### DIFF
--- a/django_filters/filterset.py
+++ b/django_filters/filterset.py
@@ -174,6 +174,7 @@ class FilterSetMetaclass(type):
         except NameError:
             # We are defining FilterSet itself here
             parents = None
+        orig_attrs = copy.copy(attrs)
         declared_filters = get_declared_filters(bases, attrs, False)
         new_class = super(
             FilterSetMetaclass, cls).__new__(cls, name, bases, attrs)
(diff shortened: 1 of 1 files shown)


django_filters/filterset.py
```diff
@@ -220,7 +220,7 @@ def __new__(cls, name, bases, attrs):
         # TODO: remove with deprecations
         # check key existence instead of setdefault - prevents unnecessary filter construction
         order_by_field = new_class._meta.order_by_field
-        if opts.order_by and order_by_field not in filters:
+        if opts.order_by and order_by_field not in attrs:
```
blueyed: That's wrong: the filter gets popped from `attrs` in `get_declared_filters`.

blueyed: `declared_filters` contains the parent's generated OrderingFilter however.

So we have to either copy `attrs` for this and/or use another method to mark the automatic OrderingFilter somehow.

blueyed: I've pushed a fixup that copies `args`.

rpkilby:
Hi @blueyed - this doesn't seem worth fixing since the option is being deprecated in the next release.

blueyed:
@rpkilby 
Ok.
What about https://github.com/carltongibson/django-filter/issues/518 then?  I am using a method to be removed there.  Closing this PR here.

blueyed:
What about having this fixed in the `0.15.x` series though?
Please feel free to close it again, but it seems to make sense, until 1.0 is finished.

rpkilby:
> What about #518 then? 

Currently responding to #518 :)

> What about having this fixed in the 0.15.x series though?

I'm not sure if there's a plan for another 0.15.x release. 0.15.1 was necessary to fix some fairly urgent issues.

rpkilby:
Given the response in #518 and the upcoming 1.0 release, is this still relevant?

PR #540, opened 2016-11-02
Fix formatting issues w/ numeric range docs

rpkilby:
Reference #343. This should resolve the documentation issues.


rpkilby:
@carltongibson - doc example added for #517.

rpkilby:
Provisional docs: http://rpkilby.github.io/django-filter/ref/filters.html#adding-custom-filter-choices

PR #546, opened 2016-11-04
Fix OrderingFilter label translation

rpkilby:
Supersedes and resolves #543.

This also adds descending options to `OrderingFilter.field_labels`.

### DIFF
--- a/django_filters/filters.py
+++ b/django_filters/filters.py
@@ -656,7 +656,7 @@ class OrderingFilter(BaseCSVFilter, ChoiceFilter):
 
     def build_choices(self, fields, labels):
         ascending = [
-            (param, labels.get(field, pretty_name(param)))
+            (param, labels.get(field, _(pretty_name(param))))
             for field, param in fields.items()
         ]
         descending = [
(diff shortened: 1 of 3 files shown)


rpkilby:
Thanks @ad-m!

django_filters/filters.py
```diff
@@ -656,7 +656,7 @@ def normalize_fields(cls, fields):
 
     def build_choices(self, fields, labels):
         ascending = [
-            (param, labels.get(field, pretty_name(param)))
+            (param, labels.get(field, _(pretty_name(param))))
             for field, param in fields.items()
         ]
         descending = [
```
ad-m: Use:
```
        descending = [
            ('-%s' % pair[0], labels.get('-%s' % pair[0], self.descending_fmt % pair[1]))
            for pair in ascending
        ]
```
to support translation of label for descending options (eg, '-username'). There is no reason to not support them, but in some cases can improve UX experience by naming orders in more natural ways.

rpkilby: That's not a bad point. It's also a fairly trivial thing to add. Currently updating the PR to include this & your corresponding test.

ad-m:
LGTM! 👍

PR #563, opened 2016-11-16
Include filter fields' help text as coreapi descriptions

unklphil:
This will make it a lot simpler to get the parameter descriptions when using DRF and Swagger, as it gets fed in automatically.


django_filters/rest_framework/backends.py
```diff
@@ -98,6 +98,7 @@ def get_schema_fields(self, view):
         filter_class = self.get_filter_class(view, view.get_queryset())
 
         return [] if not filter_class else [
-            compat.coreapi.Field(name=field_name, required=False, location='query')
-            for field_name in filter_class().filters.keys()
+            compat.coreapi.Field(
+                name=field_name, required=False, location='query', description=str(field.field.help_text))
```
unklphil: Good point! We switched to python3 two months ago and I have already forgotten about the whole unicode/str thing.

PR #580, opened 2016-12-06
Fix template packaging

rpkilby:
Starting work on #578.


rpkilby:
This should be ready.

PR #688, opened 2017-04-14
Fix for #687

DeKaN:


django_filters/rest_framework/backends.py
```diff
@@ -93,5 +93,5 @@ def get_schema_fields(self, view):
                     description=six.text_type(field.field.help_text)
                 )
             )
-            for field_name, field in filter_class().filters.items()
+            for field_name, field in filter_class(request=view.request).filters.items()
```
rpkilby: It would be better to use `request=getattr(view, 'request', None)`.

DeKaN: Changed. This attribute always exists for schema - https://github.com/encode/django-rest-framework/blob/master/rest_framework/schemas.py#L391

rpkilby:
As a general note, it's useful to add a test case that demonstrates the desired behavior. 

- This helps ensure that the patch actually fixes the problem
- It helps to prevent the feature from being lost in a potential future patch.

PR #691, opened 2017-04-19
Fix schema generation w/ request callables

rpkilby:
Alternative to #688

### DIFF
--- a/django_filters/rest_framework/backends.py
+++ b/django_filters/rest_framework/backends.py
@@ -78,7 +78,7 @@ class DjangoFilterBackend(object):
         if filter_class is None:
             try:
                 filter_class = self.get_filter_class(view, view.get_queryset())
-            except:
+            except Exception:
                 warnings.warn(
                     "{} is not compatible with schema generation".format(view.__class__)
                 )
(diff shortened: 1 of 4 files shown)


django_filters/rest_framework/backends.py
```diff
@@ -90,8 +90,8 @@ def get_schema_fields(self, view):
                 required=False,
                 location='query',
                 schema=compat.coreschema.String(
-                    description=six.text_type(field.field.help_text)
+                    description=six.text_type(field.extra.get('help_text', ''))
                 )
             )
-            for field_name, field in filter_class().filters.items()
+            for field_name, field in filter_class.base_filters.items()
```
rpkilby: FilterSet instantiation (and filters deepcopy) should be unnecessary.

django_filters/rest_framework/backends.py
```diff
@@ -90,8 +90,8 @@ def get_schema_fields(self, view):
                 required=False,
                 location='query',
                 schema=compat.coreschema.String(
-                    description=six.text_type(field.field.help_text)
+                    description=six.text_type(field.extra.get('help_text', ''))
```
rpkilby: This should be fine since form fields don't provide their own `help_text`. Technically they can, but in practice they don't.

PR #702, opened 2017-05-02
Add support for validation error codes

rpkilby:
This is for both the base `FilterSet` and the DRF variant. 

- The base `FilterSet` is slightly broken, since it simply raises `ValidationError(self.form.errors)`. This loses the individual validation error codes. 
- The `raw_validation` approach needs to be modified to return an actual `serializers.ValidationError`, since the plain data structure lacks the error codes.

(This is WIP)

### DIFF
--- a/django_filters/utils.py
+++ b/django_filters/utils.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 import warnings
 
 import django
@@ -8,7 +9,6 @@ from django.db.models.constants import LOOKUP_SEP
 from django.db.models.expressions import Expression
 from django.db.models.fields import FieldDoesNotExist
 from django.db.models.fields.related import ForeignObjectRel, RelatedField
-from django.forms import ValidationError
 from django.utils import six, timezone
 from django.utils.encoding import force_text
 from django.utils.text import capfirst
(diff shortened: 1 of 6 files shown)


rpkilby:
I'm currently stuck on this. I'm 95% sure there's a bug preventing DRF `ValidationError`s from having custom error codes. In essence, there's [this section](https://github.com/encode/django-rest-framework/blob/3.6.2/rest_framework/exceptions.py#L25-L31) which interacts with how `ValidationError`s coerce their string messages to a list [here](https://github.com/encode/django-rest-framework/blob/3.6.2/rest_framework/exceptions.py#L135). In short, instead of getting an `ErrorDetail` with a string and code, you just get a list of strings.

I need to look into this further.

rpkilby:
Note to self... ([related discussion](https://github.com/encode/django-rest-framework/issues/4787#issuecomment-326062971))

django_filters/utils.py
```diff
@@ -231,22 +232,15 @@ def label_for_filter(model, field_name, lookup_expr, exclude=False):
...
-    else:
-        return error
+    # it's necessary to lazily import the exception, as it can otherwise create
+    # an import loop when importing django_filters inside the project settings.
+    from rest_framework.serializers import ValidationError
+
+    return ValidationError(OrderedDict([
+        (key, [e.message for e in error_list])
+        for key, error_list in error.error_dict.items()
+    ]))
```
blueyed: You might want to keep codes here:
```python
        error_dict = OrderedDict([
            (key, [exceptions.ErrorDetail(e.message,
                                          e.code if e.code else default_code)
                   for e in error_list])
            for key, error_list in error_dict.items()])
        drf_validation_error = exceptions.ValidationError(error_dict)
```

blueyed:
See https://github.com/encode/django-rest-framework/issues/4787#issuecomment-337687652 for an updated version that also handles params.

rpkilby:
I'm closing this off in favor of #788, which is generally reworking filterset validation.

PR #734, opened 2017-05-31
Deprecate 'extra' field attrs on Filter

rpkilby:
I was working on some related code and it occurred to me that filters don't generally make use of the `required` and `widget` attributes. It doesn't necessarily seem beneficial to store them separately from the `extra` form field kwargs.


rpkilby:
rebased

PR #762, opened 2017-08-22
Add isort to CI & re-add warnings 

rpkilby:
- Adds `isort` tox env.
- Fleshes out the isort configuration with first/third party packages (mostly copied from DRF)
  - django and rest_framework will now precede django_filters imports
- Re-adds `warnings` job to CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,13 @@ script:
   - coverage erase
   - detox
 
+matrix:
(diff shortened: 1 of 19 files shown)


rpkilby:
This should be good to go. Given that modules are partially resorted again, I'll wait for this to merge before rebasing the other PRs.

Issue #740, opened 2017-06-21
Allow to override retrieving of a queryset in a FilterSet


Alexx-G:
The title may be a little bit confusing. It would be nice to have a generic `filter` method which is called everytime queryset is retrieved and filtering for fields is applied. Yeah, I can override `qs` property, firstly get filtered queryset and then apply prefetches. But _it's not clear if this may have some side-effects and the data passed to filter is safe to access._

There's my use case.
I'm filtering a queryset by a M2M relationship (as in `queryset.filter(items__field="foo")`). However I want to get prefetched items filtered by that field as well.

In other words my filter looks like this:
```python
# In my filterset
def filter_items_by_field(self, queryset, name, value):
        queryset = queryset.all().filter(
            items__field=value,
        )
        prefetched_items = Prefetch(
            'items',
            Item.objects..filter(
                field=value,
            ),
        )
        return queryset.prefetch_related(prefetched_items)

# In my vieset
def get_queryset(self):
        queryset = super().get_queryset()
        # This should be applied everytime
        prefetched_items = Prefetch(
            'items',
            Item.objects..filter(
                another_field=some_value,
            ),
        )
        return queryset.prefetch_related(prefetched_items)
```

A single filter like this works as expected. However, I may have several `filter_items_by_*` filters + I need to apply same filtering for queryset from prefetch for all filters. And multiple filters aren't a problem, because they're mutually exclusive in my case. But a filter + generic will cause an exception.

The point is that I can apply only one prefetch for the same lookup. So, in my case I'll simply get an exception.

I have a temporary workaround for this problem: all `filter_items_by_*` performs mandatory filtering on prefetch.
And in `filter_queryset` (after filtering was applied), if `items` aren't prefetched (what is bad since I'm inspecting queryset protected attrs), I apply prefetch with the mandatory filtering. But there should be a better way to deal with situations like this.

I hope my example is clear enough.

rpkilby:
Hi @Alexx-G. I think this could be possible with some coordination between the filterset and relevant filter classes. Essentially, the filter would append a `Prefetch` instance to its `parent` filterset. `FilterSet.qs` would then simply call `prefetch_related` when appropriate.

That said, what's the use case for this? I would argue that there are likely more appropriate places to handle prefetching.

Alexx-G:
>  I think this could be possible with some coordination between the filterset and relevant filter classes. 

[...]

PR #783, opened 2017-09-24
Extract form validation and queryset filtering from caching

rpkilby:
I was thinking about #740 a while back, and realized that overriding the `FilterSet.qs` property is slightly awkward, due to the `_qs` caching. In the case of #740, `Prefetch` calls can only occur once, as duplicate calls can raise an exception. If you want to cache the prefetch, a correct solution looks like the following:

```python
@property
def qs(self):
    # ensure that we haven't already cached, as this also implies prefetching
    already_prefetched = hasattr(self, '_qs')
    super().qs
    if self.is_bound and not already_prefetched:
        self._qs = self._qs.prefetch_related(...)
    return self._qs
````

Overriding `filter_queyset()` is a little more intuitive, as it would only be executed once.

```python
def filter_queryset(self, queryset):
    qs = super().filter_queryset(queryset)
    return qs.prefetch_related(...)
```

### DIFF
--- a/django_filters/filterset.py
+++ b/django_filters/filterset.py
@@ -214,27 +214,28 @@ class BaseFilterSet(object):
         if not hasattr(self, '_qs'):
             if not self.is_bound:
                 self._qs = self.queryset.all()
-                return self._qs
-
-            if not self.form.is_valid():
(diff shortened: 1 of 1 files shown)


rpkilby:
Temporarily closing in favor of alternative approach.

PR #800, opened 2017-10-20
Method signature improvements

rpkilby:
- Rename `f` => `field` in FilterSet classmethods
- Reorder `Filter.__init__` args to be mostly keyword-only (depends on going py3-only).


rpkilby:
Give me a sec to resolve the conflicts

PR #828, opened 2017-11-17
Add submit button on crispy

vandersondev:
Add submit button on crispy form template, and change crispy tag for crispy filter

### DIFF
--- a/django_filters/templates/django_filters/rest_framework/crispy_form.html
+++ b/django_filters/templates/django_filters/rest_framework/crispy_form.html
@@ -2,4 +2,7 @@
 {% load i18n %}
 
 <h2>{% trans "Field filters" %}</h2>
-{% crispy filter.form %}
+<form class="form" action="" method="get">
+    {{ filter.form|crispy }}
+    <button type="submit" class="btn btn-primary">{% trans "Submit" %}</button>
+</form>


rpkilby:
Hi @vandersondev, does the current implementation not have a submit button? I was under the impression that the crispy tag used the layout helper, which provides the button (see: [here](https://github.com/carltongibson/django-filter/blob/master/django_filters/rest_framework/filterset.py#L25-L37)).

vandersondev:
See

image

My version is 1.1.0 (I installed with pip)

So, I putted a ipdb in FilterSet.form method, but the execution dont stop at the break point

image


I changed the files with the git repository version, but...

http://dpaste.com/0MZCPNE

And that is why I put the button hard coded as in django_filters/rest_framework/form.html


Sorry my bad english

rpkilby:
Hi @vandersondev - I would guess this has something to do with your setup. This is what I'm seeing:

screen shot 2017-11-20 at 10 14 37 pm

What `FilterSet` class are you using? Are you using `django_fitlers.FilterSet` or are you using `django_filters.rest_framework.FilterSet`?

vandersondev:
Oh, sorry.

I misunderstood the integration between django filters and DRF.
I'm reading Lightweight Django, but had a lot of inconsistencies.

rpkilby:
As far as I understand it, Lightweight Django is a great resource. Unfortunately, it was written in 2014 when Django 1.6 and 1.7 were the latest releases, so it's a bit outdated.

carltongibson:
My favourite bit was the minimal django setup example. There's another version of that here: https://github.com/syntarsus/minimal-django

PR #851, opened 2018-01-15
Rework 'lookup types' handling into new `LookupChoiceFilter` class

### LINKED ISSUE #843: Build fails against Django (pre-2.1) master
Hello,

Just as a heads up, your build is failing against the latest version of master in the Django repo, because they removed `from django.db.models.sql.constants import QUERY_TERMS`.

Apologies if it's a bit too early for that or if you already know it.

rpkilby:
Remove "lookup types" handling for `Filter.lookup_expr`, where a `list` or `None` value would trigger the use of `LookupTypeField`. This has been replaced with a dedicated `LookupChoiceFilter`. 

I decided to replace the feature outright and raise an assertion with the relevant removal message, instead of attempting to go through a deprecation path. While it should be possible, it didn't seem worth the effort to try support/deprecate the old functionality, especially given the clean break 2.0 provides.

Should also fix #843 (at least the reliance on the defunct `QUERY_TERMS`). 

**Changes:**
- Add `LookupChoiceFilter`
- Remove `list` and `None` value handling for `lookup_expr`. Using this form will fail an assertion check. 
- Remove `LOOKUP_TYPES` module variable from `django_filters.filters`.
- Rename `LookupTypeWidget` to `LookupChoiceWidget`.
- Rename `LookupTypeField` to`LookupChoiceField`.
- Rename `Lookup.lookup_type` to `Lookup.lookup_expr`.
- Change `Lookup` to only accept non-empty values. If a `Lookup` is missing a value, a `None` should be provided instead.


rpkilby:
Yep - this does fix #843, in so far as the build no longer fails to even run 😄

PR #855, opened 2018-01-16
Fix distinct call for range filters

rpkilby:
Per @joinworldtv with https://github.com/carltongibson/django-filter/pull/417#issuecomment-357722081

Instead of reimplementing the `.filter()` method entirely, the override simply alters the `.lookup_expr` and value, then calls the base method.

### DIFF
--- a/django_filters/filters.py
+++ b/django_filters/filters.py
@@ -390,16 +390,15 @@ class NumericRangeFilter(Filter):
     def filter(self, qs, value):
         if value:
             if value.start is not None and value.stop is not None:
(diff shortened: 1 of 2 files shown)


rpkilby:
> So surely there's a missing test case here? :slightly_smiling_face: 

I was thinking about that. The updated `.filter()` method is simply altering the lookup expression and value, before calling the base implementation. The current tests sufficiently cover that behavior, so it didn't seem necessary to add specific tests for compatibility with the other features (like `exclude` and `distinct`). 

That said a sanity check wouldn't hurt.

rpkilby:
Okay, updated with some sanity tests for calling `distinct`.

PR #884, opened 2018-03-09
Support for PostgreSQL to_tsvector function with multiple fields

Charnelx:
Hi.
I added SearchVectorFilter to support to_tsvector function for full text search. Theare two problems I faced:

- this thing is vendor-specific (yep, only PostgreSQL)
- to search by multiple fields we need to create one more argument (e.x. field_names) or use existing field_name with additional type check

Pros:
- fast search
- search by multiple fields at one's

Cons:
- used only with PostgreSQL
- dual meaning of field_name argument wich may lead to misunderstanding and mistakes

Case of use:

**models.py**
```python

class Person(models.Model):

    name = models.CharField(max_length=128, verbose_name='person name')
    surname = models.CharField(max_length=128, verbose_name='person surname')
```
**views.py**

```python
class PersonFilter(filters.FilterSet):
    person = SearchVectorFilter(field_names=['name', 'surname'], lookup_expr='icontains')
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,7 +20,7 @@ Requirements
 ------------
 
 * **Python**: 3.4, 3.5, 3.6
-* **Django**: 1.11, 2.0b
+* **Django**: 1.11, 2.0
 * **DRF**: 3.7
 
 If you need to support Python 2.7 or Django <1.11 use the version 1.1 release.
--- a/docs/guide/install.txt
+++ b/docs/guide/install.txt
@@ -30,5 +30,5 @@ __ http://www.django-rest-framework.org/
 
 
 * **Python**: 3.4, 3.5, 3.6
-* **Django**: 1.11, 2.0b
+* **Django**: 1.11, 2.0
 * **DRF**:  3.7
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,7 @@ setup(
         'Operating System :: OS Independent',
         'Framework :: Django',
         'Framework :: Django :: 1.11',
+        'Framework :: Django :: 2.0',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.4',
--- a/tox.ini
+++ b/tox.ini
@@ -18,7 +18,7 @@ setenv =
        PYTHONDONTWRITEBYTECODE=1
 deps =
     django111: django>=1.11.0,<2.0
-    django20: django>=2.0b1,<2.1
+    django20: django>=2.0,<2.1
     djangorestframework>=3.7,<3.8
     latest: {[latest]deps}
     -rrequirements/test-ci.txt


django_filters/utils.py
```diff
@@ -78,8 +78,19 @@ def get_field_parts(model, field_name):
         >>> [p.verbose_name for p in parts]
         ['author', 'first name']
 
+        >>> parts = get_field_parts(Book, [author__first_name, 'surname'])
+        >>> [p.verbose_name for p in parts]
+        ['author', 'first_name', 'surname']
```
rpkilby: `get_field_parts` is not intended to return a list of field names, but a list of the *individual parts* of a related field name (FK, m2m, 1to1). This result would be nonsensical, as `first_name` is a regular field on the author model, and does not represent a relationship to an additional `Name` model or something. 

Either way, as explained above, overloading `field_name` is not the correct path, and these changes are a symptom of that. I would revert these changes.

django_filters/filters.py
```diff
@@ -484,6 +486,38 @@ class TimeRangeFilter(RangeFilter):
     field_class = TimeRangeField
 
 
+class SearchVectorFilter(Filter):
```
rpkilby: I would inherit `CharFilter`, as you are validating text.

django_filters/filters.py
```diff
@@ -484,6 +486,38 @@ class TimeRangeFilter(RangeFilter):
...
+
```
rpkilby: Instead of overloading the purpose of `field_name`, create a custom `__init__` that requires users to pass in a list of search fields. eg, 

```python
# Note that this provides a default `field_name` to be used for the annotation
# and query, and that `search_fields` will be a required keyword argument.
def __init__(self, field_name='search_vector', lookup_expr='exact', *, search_fields, **kwargs):
    super().__init__(field_name, lookup_expr, **kwargs)

    self.search_fields = search_fields
```

From here, it's really simple to overload the `.filter()` method to add in the annotation

```python
def filter(self, qs, value):
    # Note that the annotation is operation on the filter's `field_name`, which was
    # set above to 'search_vector' as the filter's default value.
    qs.annotate(self.field_name=SearchVector(self.search_fields)
    
```

django_filters/utils.py
Charnelx: You truth, I have read method descption about "the traversable relationships" and..ignored it.

django_filters/filters.py
Charnelx: Reasonably and thus accepted.

django_filters/filters.py
Charnelx: Yes, this was the way it should be implemented. Thanks.

Charnelx:
@rpkilby great thanks for detailed response!
Well, adding additional argument to Filter was my first idea but I thought that I would be able to make something that not require any external change at all. Now I see the failure of this approach.

I'll take into account your remarks and make another try + will write tests.

Issue #754, opened 2017-08-02
Proposal for getting help_text from model fields


jter:
It would be great that filters of `FilterSet` could get `help_text` from fields of Models. This feature can save some repetitive work when we need `help_text` for auto-generated docs from model fields without redefining the filters in `FilterSet`. Could this proposal be a good idea?

rpkilby:
Hi @jter. I feel like this has been talked about before, but I can't find the relevant discussion. In short, this suggestion works well for FilterSets like...

```python
class UserFilter(FilterSet):
    class Meta:
        model = User
        fields = ['username', 'email', 'name', ...]
```

Adding the `help_text` is reasonable in this case, since the filter's data type matches the data type of the underlying model field. eg `/api/users?username=jter`

However, `Meta.fields` can also be used like so:

```python
class ArticleFilter(FilterSet):
    class Meta:
        model = Article
        fields = {
            'author': ['exact', 'startswith', 'endswith', ...],
            'title': ['exact', 'startswith', 'endswith', ...],
            'published_at': ['exact', 'isnull'],
        }
```

In this case, adding in the model's `help_text`s is a little less useful.

1. There may be several filters for the same model field, making the `help_text` repetitive/redundant.
2. The data type of the filter doesn't necessarily match the data type of the model field. eg, information on entering a datetime may not be appropriate for an `isnull` lookup where the input is a boolean. 

----

This would be fairly straightforward to achieve with a custom FilterSet base class:

```python

class HelpfulFilterSet(django_filters.FilterSet):
    @classmethod
    def filter_for_field(cls, f, name, lookup_expr):
        filter = super(HelpfulFilterSet, cls).filter_for_field(f, name, lookup_expr)
        filter.extra['help_text'] = f.help_text
        return filter
```

jter:
Thank you for your illustrating, clear and useful.
@rpkilby

PR #891, opened 2018-04-04
Add 'help_text' django model field option support

2miksyn:
It useful, for example, in coreapi documentation of DRF


rpkilby:
Hi @2miksyn. The `help_text` is omitted intentionally. #754 explains why this is the case.

2miksyn:
Facts sounds good. Thinks moments of inheritance `help_text` from model would be nice to add in docs.

PR #901, opened 2018-05-02
An empty `data` can result in an empty queryset (even if data exists)

claytondaley:
The attached shows a behavior that strikes me as counter-intuitive but may be intentional.  

It's my assumption that a FilterSet that receives no `data` should not apply any filters *whether or not a queryset is provided*.

 - In the previous assertion (of the same test), this is verified for `data=None` and `queryset=None`.
 - I've added a version of the test where `queryset` is not None (i.e. the caller starts by providing `all()`).

The test now fails because `F(data=None, queryset=User.objects.all())` actually returns `User.objects.none()`.  This happens because:

 - `CharFilter` depends on Django's `forms.CharField`
 - `forms.CharField` defaults to `''` for the value `None` ([as of 1.11](https://docs.djangoproject.com/en/2.0/ref/forms/fields/#django.forms.CharField.empty_value), this value is configurable, but it only made explicit the old hard-coded behavior)
 - the `qs` property on `BaseFilterSet` only ignores the filter [if the cleaned value `is None`](https://github.com/carltongibson/django-filter/blob/1.1.0/django_filters/filterset.py#L219)

### DIFF
--- a/tests/test_filtering.py
+++ b/tests/test_filtering.py
@@ -1160,6 +1160,7 @@ class FilterMethodTests(TestCase):
                 return queryset.filter(**{name: value})
 
         self.assertEqual(list(F().qs), list(User.objects.all()))
+        self.assertEqual(list(F(queryset=User.objects.all())), list(User.objects.all()))
         self.assertEqual(list(F({'username': 'alex'}).qs),
                          [User.objects.get(username='alex')])
         self.assertEqual(list(F({'username': 'jose'}).qs),


tests/test_filtering.py
```diff
@@ -1160,6 +1160,7 @@ def filter_username(self, queryset, name, value):
                 return queryset.filter(**{name: value})
 
         self.assertEqual(list(F().qs), list(User.objects.all()))
+        self.assertEqual(list(F(queryset=User.objects.all())), list(User.objects.all()))
```
rpkilby: btw, the filterset is no longer directly iterable. This was removed in #440

`F(queryset=User.objects.all())` => `F(queryset=User.objects.all()).qs`

claytondaley:
You're right.  Fixing `qs` fixed this test and adding `data={}` didn't recreate it.  I'm seeing something equivalent IRL so I'll keep massaging until I figure out the actual condition.

rpkilby:
Are you getting no results on a `GET` request without any query params? Even when there are no query params provided, an empty `data` dict is still bound to the FilterSet.

claytondaley:
The empty value *is* getting caught in the next layer (i.e. the `filter` call) so the `FilterSet` may not be responsible.  I traced the issue down to `get_object` and I thought I had traced it into `queryset = self.filter_queryset(self.get_queryset())` using a watch.  When I created a test that seemed to reproduce, I stopped searching.

rpkilby:
If you're using the default backend, then note that the filterset is always bound with the `query_params`.

https://github.com/carltongibson/django-filter/blob/af6cff84a2ff3e09e94e4b8649cd06a417dd5228/django_filters/rest_framework/backends.py#L50-L56

claytondaley:
Yes.  It looks like it's actually getting past this call so I definitely shouldn't waste more of your time.

rpkilby:
Good luck with the bug hunt.

PR #920, opened 2018-05-30
DRF backend: view.get_filter_class support

rqblast:
In my case, I wanted to select different filter classes depending on the action / method in the ViewSet.
For example, `if self.action == 'list'` then I want to select some `ListFilter` in another cases - some `DefaultFilter`.
I think this ability can be useful with ViewSets.

### DIFF
--- a/django_filters/rest_framework/backends.py
+++ b/django_filters/rest_framework/backends.py
@@ -20,7 +20,10 @@ class DjangoFilterBackend(object):
         """
         Return the django-filters `FilterSet` used to filter the queryset.
         """
+        get_filter_method = getattr(view, 'get_filter_class', None)
         filter_class = getattr(view, 'filter_class', None)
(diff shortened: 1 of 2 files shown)


rpkilby:
Hi @rqblast. The more appropriate place to open this issue is in django-filter. 

That said, you should be able to implement this with a property on the view. 

```python

class MyViewSet(...):
    @property
    def filter_class(self):
        if self.detail:
            return MyDetailFilterSet
        else:
            return MyListFilterSet
```

Note that you can use the relatively new `ViewSet.detail` initkwarg to switch between list/detail variants.

rqblast:
Hi @rpkilby 
Thanks for the trick. Can I close this PR?

rpkilby:
Probably. I don't think expanding the API surface is hugely beneficial, given that the problem can be solved by using a property.

PR #923, opened 2018-06-07
Add `renderer` argument to `render()` method of `BooleanWidget`

bronsen:
This ought to help against:
> .../lib/python3.6/site-packages/django/forms/boundfield.py:36:
> RemovedInDjango21Warning: Add the `renderer` argument to the render() method
> of <class 'django_filters.widgets.BooleanWidget'>. It will be mandatory in
> Django 2.1.

The renderer argument was added in Django 1.11: https://docs.djangoproject.com/en/2.0/ref/forms/widgets/#django.forms.Widget.render

I have seen this deprecation warning only in association with the BooleanWidget, so only this class was modified.

### DIFF
--- a/django_filters/widgets.py
+++ b/django_filters/widgets.py
@@ -23,7 +23,7 @@ class LinkWidget(forms.Widget):
         self.data = data
         return value
 
-    def render(self, name, value, attrs=None, choices=()):
+    def render(self, name, value, attrs=None, choices=(), renderer=None):
         if not hasattr(self, 'data'):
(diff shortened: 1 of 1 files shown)


michael-k:
> I have seen this deprecation warning only in association with the BooleanWidget

@bronsen I guess that's because we don't use the other affected widgets (`LinkWidget` and `BaseCSVWidget`).

bronsen:
So it would make sense to modify the other widgets as well. Good, I shall do so!

PR #937, opened 2018-07-03
Test QueryDict against CSV filters

rpkilby:
Verify expected behavior from #935. Essentially, the CSV filters should treat:
```
MyAPI/mymodel?mysubmodel__in=1&mysubmodel__in=5&mysubmodel__in=9
```

as equivalent to:

```
MyAPI/mymodel?mysubmodel__in=9
```

- Refactors tests to use Python 3's `subTest` feature. 
- Updates tests to test against expected IDs instead of just a count.
- Adds a QueryDict case to test the above expected behavior.

### DIFF
--- a/tests/test_filtering.py
+++ b/tests/test_filtering.py
@@ -4,6 +4,7 @@ import unittest
 from operator import attrgetter
 
 from django import forms
+from django.http import QueryDict
 from django.test import TestCase, override_settings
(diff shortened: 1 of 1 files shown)


rpkilby:
> Why do they take the last value (rather than the first say?)

I honestly have no idea.

PR #938, opened 2018-07-04
Prevent having queryset=None

jpic:
This will forgive user errors where their filter() implementation
returns None, ie. in the case of a choice filter where they don't handle
the case where no choice was selected.

Otherwise, we could raise a clear Exception, because currently in this case users would get "NoneType has no attribute filter" which might not help the user understand how they should fix their issue. 

TYIA for comments

### DIFF
--- a/django_filters/filterset.py
+++ b/django_filters/filterset.py
@@ -178,7 +178,9 @@ class BaseFilterSet(object):
         applied to the queryset before it is cached.
         """
         for name, value in self.form.cleaned_data.items():
-            queryset = self.filters[name].filter(queryset, value)
+            result = self.filters[name].filter(queryset, value)
+            if result is not None:
+                queryset = result
         return queryset
 
     @property


rpkilby:
I would be -1 on accepting `None` values as this is most likely the result of user error, and silently accepting this would just lead to confusion (Why isn't my filter being called? Why is django-filter broken?). That said, asserting that the result is a `QuerySet` does seem like a good idea. 

Also, a test demonstrating the behavior would be great.

jpic:
@rpkilby most tests already demonstrate this behaviour. 

Raising an exception here will break most of them, and require a rewrite of them.

Currently, it seems for me that "NoneType has no attribute filter" is most confusing, how do you recommend to solve that ?

rpkilby:
Hi @jpic. Sorry if I wasn't clear - I was asking those questions as a hypothetical user, the idea being that accepting `None` as valid may be confusing. For example, I've created some hypothetical custom filter method, but mistakenly leave off the return statement. When I attempt to use the filter, what happens?

With the current implementation, it breaks via the unexpected `AttributeError`. Unfortunately, the error isn't as helpful as it could be, since it doesn't indicate which filter is broken. However, you at least know that something *is* broken.

With the PR, seemingly nothing happens, as the implicit `None` return value is ignored. You know something isn't right, because you don't see a change in the result, but nothing is explicitly complaining either. This break in expectations is where a user might ask "Why isn't my filter being called? I don't see anything happening." or "Why is django-filter broken?". While the current exception isn't incredibly helpful, it at least points you to where something is breaking, instead of obscuring it. 

That's why I was recommending a more explicit assertion. It allows us to preempt the attribute error and provide more meaningful information, such as the filter's attribute name and that a queryset is the expected result. 

> most tests already demonstrate this behaviour (hence, coverage + 0.01%). Raising an exception instead breaks most of them, requiring a rewrite.

Sure, but the behavior is tested incidentally, not explicitly. Coverage only went up because the tests already walk these code paths. If the nature of those tests change, this behavior would (potentially) no longer be tested.

> Can you recommend an alternative fix ? I'm willing to contribute any kind of code to prevent the current misleading behavior.

Honestly, the only thing that makes sense in my mind is to use an assertion. I guess the question is whether or not the improved exception is worth updating the tests.

jpic:
> However, you at least know that something is broken.

Well there's the case where you don't want to apply any particular filter, ie. value of filter is None. But we could consider this case marginal.

> Honestly, the only thing that makes sense in my mind is to use an assertion. I guess the question is whether or not the improved exception is worth updating the tests.

Are you in favor of leaving it as is ?

Thank you for your reply but it's still unclear to me. I will close this PR and if you want to create an issue and assign it to me feel free.

PR #943, opened 2018-07-13
Improve exception message for invalid filter result

rpkilby:
As mentioned in #938, the exception for an invalid filter result isn't that helpful. e.g., if a custom filter method unintentionally returns a `None` value, you end up with `NoneType has no attribute filter`, which lacks any context about the offending `FilterSet` and `Filter` instance.

@jpic - would this exception have been helpful the first time around? e.g.
```
Expected 'F.invalid' to return a QuerySet, but got a NoneType instead.
```

### DIFF
--- a/django_filters/filterset.py
+++ b/django_filters/filterset.py
@@ -179,6 +179,9 @@ class BaseFilterSet(object):
         """
         for name, value in self.form.cleaned_data.items():
             queryset = self.filters[name].filter(queryset, value)
+            assert isinstance(queryset, models.QuerySet), \
+                "Expected '%s.%s' to return a QuerySet, but got a %s instead." \
+                % (type(self).__name__, name, type(queryset).__name__)
         return queryset
 
     @property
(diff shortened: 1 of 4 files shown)


jpic:
LGTM

stefanw:
I would like to use `django-filter` for things besides Django's querysets like Haystack's `SearchQuerySet`. The assertion in this PR with an `isinstance` (not even `issubclass`) is a bit too assuming. Maybe a `None` check would suffice to report a common error?

rpkilby:
Hi @stefanw. django-filter isn't going to be compatible with `SearchQuerySet` since a SQS is not compatible with the standard `QuerySet` API. They have similarities, but they are not compatible with each other. e.g., you cannot combine a `SearchQuerySet` with a standard `QuerySet` via `&` or `|`, `SearchQuerySet` cannot accept `Q` objects, etc...

The `filter_queryset` method is functionally 3 lines of code. If you want django-haystack compatibility, I would recommend creating a custom base filterset class that overrides this method.

PR #1005, opened 2018-11-13
Add IsoDateTimeFromToRangeFilter (#1004)

jobec:
Add `IsoDateTimeFromToRangeFilter` similar to `DateTimeFromToRangeFilter`


jobec:
I only now see #1003 which is basically the same functionality as this pull request. 😊

Issue #1013, opened 2018-12-10
Meta.fields dict usage silently fails on declared filters


rpkilby:
The below is a simplified version of https://stackoverflow.com/q/53411473/1103124

```python
class MyModel(models.Model):
    name = models.CharField()


class CustomFilter(django_filters.FilterSet):
    value = django_filters.CharFilter(field_name="name")

    class Meta:
        model = MyModel
        fields = {
            'value': ['exact', 'iexact', 'contains', 'icontains', ...],
        }
```

In the above, the user has aliased the model's `name` field to `value`. They then try to generate additional filters against the parameter name instead of the underlying model name, but this fails silently. This should be raising `TypeError("'Meta.fields' contains fields that are not defined on this FilterSet: value")`.

The underlying issue is the check here:
https://github.com/carltongibson/django-filter/blob/15683cd14e0f20a0bed8abab3448b5789e78fc62/django_filters/filterset.py#L348

Since `value` is a declared filter, it's removed from the list. 

The "missing fields" handling should be reworked to "missing field/lookup expressions". I assume this code also silently fails for non-existent lookups such as `foo`.

----

One alternative solution would be to support #774. It would be possible to use the declared filter's attribute name and `field_name` as the basis for generating additional aliased filters, but I have a feeling this could be too magical/complex. e.g., would users expect the declared filter's arguments (e.g., `choices`) to be copied or not?

PaiAkshay998:
Ran into this myself yesterday. The docs seemed to imply the ```fields``` tag would be used solely in conjunction with ```models```.  I've closed my PR for now because now I'm not sure about the exact functionality that the ```fields``` keyword is supposed to bring in.

rpkilby:
Having a declared field in `Meta.fields` is acceptable. The issue is that an error should be raised when trying to use it to generate extra lookups.

PaiAkshay998:
Ah alright. Yeah that makes sense.

PaiAkshay998:
I've reopened my pr. Basically I've included an extra check for all fields other than the LookupChoiceFilter to check that if a field which is specified in ```Meta.fields``` also exists in declared fields, then make sure that the lookup expressions match; ie declaration of fields in both ```Meta.fields``` and the declare fields agree on the same lookup expressions for that field.

I'm not sure if this covers all possible cases as I'm not entirely familiar with the codebase yet. Any feedback would be greatly appreciated!

rpkilby:
A couple of updates: 
[...]

PR #1019, opened 2018-12-16
Prevent declared field to have additional lookups

PaiAkshay998:
I've included an extra check for all fields other than the LookupChoiceFilter to check that if a field which is specified in Meta.fields also exists in declared fields, then make sure that the lookup expressions match; ie declaration of fields in both Meta.fields and the declare fields agree on the same lookup expressions for that field.

I'm not sure if this covers all possible cases as I'm not entirely familiar with the codebase yet. Any feedback would be greatly appreciated! #1013

### DIFF
--- a/django_filters/filterset.py
+++ b/django_filters/filterset.py
@@ -22,6 +22,7 @@ from .filters import (
     DateTimeFilter,
     DurationFilter,
     Filter,
+    LookupChoiceFilter,
     ModelChoiceFilter,
     ModelMultipleChoiceFilter,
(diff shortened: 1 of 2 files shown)


PaiAkshay998:
@rpkilby Was wondering if this was still required?

rpkilby:
Hi @PaiAkshay998. You can close this. The issue isn't directly related to `LookupChoiceFilter` (although it happens to be used in the original SO question). I've opened #1061, which fixes this more generally.

PR #1028, opened 2019-01-21
Fix indentation in docs/ref/filters.txt

jieter:
https://django-filter.readthedocs.io/en/master/ref/filters.html#datefromtorangefilter

note/warning after code block was not properly indented


jieter:
also note readthedocs still serves docs for version 2.0

PR #1059, opened 2019-03-21
add relative time example to documentation

dustinmichels:
Hi!

It took me a long time to figure out how to write a `django-filters` filter to get data based on a relative timestamp, eg past 24 hours.

I thought I would offer the solution I eventually found to the "tips" section of your documentation.

I'm not at all offended if you don't want to include this!

### DIFF
--- a/docs/guide/tips.txt
+++ b/docs/guide/tips.txt
@@ -209,6 +209,35 @@ behavior as an ``isnull`` filter.
             model = MyModel
 
 
+Filtering by relative times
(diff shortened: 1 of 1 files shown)


dustinmichels:
Not sure why checks are failing... I only changed `docs/guide/tips.txt`

rpkilby:
Hi @dustinmichels. This is an unrelated issue with one of the datetime tests. See #1050.

PR #1070, opened 2019-04-17
add help_text in filter_class.extra

dhgdhg:
add help_text in filter_class.extra to show help_text in drf docs


rpkilby:
It looks like you opened two PRs, although the other fixes a linter issue.

PR #1071, opened 2019-04-17
add help_text in filter_class.extra

dhgdhg:
add help_text in filter_class.extra to show help_text in drf docs


rpkilby:
Hi @dhgdhg. Model field help text is intentionally non included, per the reasons in  https://github.com/carltongibson/django-filter/issues/754#issuecomment-321059424. 

That said, maybe a PR adding this as a docs example would be useful?

PR #1088, opened 2019-06-11
Inclusion of the possibility of logic fields in django-filter

DavidLSO:
When we have fields that are for logical use we can declare it as follows:
```
    import django_filters

        name = django_filters.CharFilter(lookup_expr='iexact')
        show_more_details = django_filters.BooleanFilter()

        class Meta:
            model = Product
            fields = ['price', 'release_date', 'show_more_details']
            logic_fields = ['show_more_details']
```

This logic field can be used for example in the ``get_context_data``.

### DIFF
--- a/django_filters/filterset.py
+++ b/django_filters/filterset.py
@@ -221,10 +221,11 @@ class BaseFilterSet(object):
         applied to the queryset before it is cached.
         """
         for name, value in self.form.cleaned_data.items():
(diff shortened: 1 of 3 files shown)


rpkilby:
Could you explain what the purpose of this is? If you want to add fields to the form without enabling filtering for that field, you can achieve this by overriding the `Meta.form` class.

PR #1092, opened 2019-07-01
Add lookup_fields in REST Backend

muriloventuroso:
This allows you to change the lookup of the fields defined in filterset_fields in the integration with Django Rest Framework.


rpkilby:
Why don't you just use the dict form of `filterset_fields`? e.g., 

```python
class MyView(GenericAPIView):
    filterset_fields = {
        'text': ['exact'], 
        'decimal': ['exact'],
        'date': ['gt', 'lt'],
    }
    ...
```

PR #1118, opened 2019-09-02
Rework ordering filter

earshinov:
See https://github.com/carltongibson/django-filter/issues/1021


earshinov:
@carltongibson , @rpkilby , Could you take a look and share your thoughts?

I started with the code and tests.

TODO:
- [ ] document in docs/ref/filters.txt
- [ ] update docs/guide/migration.txt

Jihadik:
@earshinov, any updates here? I would be happy to use `nulls_last` feature.

earshinov:
@Jihadik , Hi! No, I don't have any updates. I think I have pushed everything that I had, now awaiting feedback from django-filter maintainers. The version from this PR should already support `nulls_last` just fine. I would appreciate if you gave it a try and shared your feedback. It seems that the maintainers are not terribly interested in this feature, and additional user feedback could convince them to pay more attention to it, should they have time.

django_filters/filters.py
```diff
@@ -655,100 +658,248 @@ def filter(self, qs, lookup):
...
+        if 'params' not in kwargs:
+            fields = kwargs.get('fields', {})
+            fields = self.normalize_fields(fields)
+            field_labels = kwargs.get('field_labels', {})
+            params = self.fields_to_params(fields, field_labels)
+        else:
+            params = self.normalize_params(kwargs['params'])
 
-        self.param_map = {v: k for k, v in fields.items()}
+        self._params = params
```
Jihadik: Looking at the code around `_params` is going against the style. Just use `self.params`

tests/models.py
```diff
@@ -41,8 +41,8 @@ def formfield(self, **kwargs):
 
 class User(models.Model):
     username = models.CharField(_('username'), max_length=255)
-    first_name = SubCharField(max_length=100)
-    last_name = SubSubCharField(max_length=100)
+    first_name = SubCharField(max_length=100, null=True, blank=True)
+    last_name = SubSubCharField(max_length=100, null=True, blank=True)
```
Jihadik: I believe, by default an empty string is null. Also, you are not using forms, so `blank=True` is not needed. hence, you don't need to touch these two lines.

Jihadik:
@carltongibson, could I get attention for this PR?
@earshinov, it would be cool if you could merge the latest master, thanks!

earshinov:
@Jihadik , I rebased the PR on top of the `master` branch

django_filters/filters.py
earshinov: Indeed, it seems that in the existing code underscores are only utilized to name backing fields used in the implementation of get/set properties. Renamed in edda682.

tests/models.py
earshinov: > I believe, by default an empty string is null

I am not sure what you mean. `null` and empty string are different values and are treated differently both by Django ORM and the underlying database.

`null=True` is necessary because within tests I create some User instances with first_name=null to check handling of nulls. The default value for the `null` constructor parameter is False.

As for `blank=True`, maybe User instances are not tied to forms at the moment, but they can be in the future. I think it's a good practice to set `blank=True` once one defines `null=True`.

Jihadik: This is from the Django official doc: https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.Field.null

earshinov: > This is from the Django official doc: https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.Field.null

Still, having `null=False` will prevent us from testing `User(first_name=null)`

> As for the blank=True, I believe it should be added on demand. Otherwise, me reading the tests would scratch my head and think why is it necessary to have it here?

Well, I would rather scratch my head if I saw a field with null=True but blank=False. It clearly won't work with forms at all. I guess we need another opinion here, ideally from django-filter devs.

gabn88: I have some experience with Django and would like to share my opinion on this matter. I think having blank=True, null=True on a django CharField is definitely not a best practice and even strongly disencouraged. It means you have two empty values in the database, NULL and '' (empty char). Therefore my recommendation is to rewrite the test by using a datetimeField (for example) with null=True, blank=True. In that case it makes logical sense and the correctness of the test is also faily easily determined.

Having said that I would also love to see this merged! Big applause for @earshinov for undertaking this project!

earshinov: Hi @gabn88! Sorry that I didn't have time to reply during workdays. Thank you for your input, your suggestion about avoiding nullable/blank CharFields seems very reasonable. I reworked the tests using a DateTime model field instead of CharField and rebased the branch onto the latest `master` along the way. Hopefully these changes will help merge this PR.

forgoty:
What is the progress about this task?
@rpkilby

jaap3:
Too bad this has gone stale, I'm in need of a way to have null values sorted to the bottom and had hoped django-filter supported it. Maybe some day...

axieum:
I'm also keen to see this across the line 🙌

PR #1125, opened 2019-09-24
added django 3.0 in tox file

b4isty:


.travis.yml
```diff
@@ -26,6 +26,8 @@ matrix:
   allow_failures:
     - env: TOXENV=warnings
 
+
```
auvipy: these are not needed.

auvipy:
https://github.com/encode/django-rest-framework/pull/6953/files

tox.ini
```diff
@@ -4,6 +4,7 @@ envlist =
        {py35,py36}-django20,
        {py35,py36,py37}-django21,
        {py35,py36,py37}-django22,
+       {py36, py37}-django30,
```
Djailla: No space between versions, keep the other line style

tox.ini
```diff
@@ -24,6 +25,7 @@ deps =
     django20: django==2.0.*
     django21: django==2.1.*
     django22: django==2.2.*
+    django30: django>=3.0a1,<3.1
```
Djailla: Try to keep the same style

like `django==3.*`

rpkilby: Unfortunately, `django==3.0.*` won't match alpha/beta pre-releases, so it's necessary to use this style until the final version is released.

PR #1129, opened 2019-10-01
Customizable default lookup expression

hodossy:
When there are mosty string fields in an application, and one does not like the `__icontains` or `__iexact` suffixes in the URL params, then it would be beneficial to change the default lookup expression generated.

### DIFF
--- a/django_filters/conf.py
+++ b/django_filters/conf.py
@@ -7,6 +7,8 @@ from .utils import deprecate
 DEFAULTS = {
     'DISABLE_HELP_TEXT': False,
 
+    'DEFAULT_LOOKUP_EXPR': 'exact',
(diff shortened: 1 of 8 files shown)


hodossy:
Thanks for the review and the helpful comments. I have added several more tests as per the suggestions.

rpkilby:
No worries - happy to help. There are a few more linter errors. You can check these with 

```
tox -e isort,lint
```

hodossy:
I think the latest commit resolved those issues

hodossy:
I think those are resolved now.

rpkilby:
Just did a re-review. Aside from the conflict this looks good to go. Ace work.

PR #1130, opened 2019-10-08
Extend test coverage

philsupertramp:
Hey,

This is my first Pull Request to an open source project so I thought it would be nice to contribute to something which I use frequently, thanks for that!
I decided to do that by improving the test coverage, but needed to leave some "holes" in `compat.py` and `__init__.py` because I don't know how to test these cases well.

Sorry for all of the small commits, I can squash them down into one if you want =)

### DIFF
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -1,7 +1,3 @@
-
-import warnings
-
-from django.conf import settings as dj_settings
 from django.test import TestCase, override_settings
 
 from django_filters.conf import (
@@ -107,4 +103,4 @@ class SettingsObjectTestCase(TestCase):
                 DEPRECATED_SETTINGS.append('TEST_123')
                 DEFAULTS['TEST_123'] = True
                 settings.change_setting('FILTERS_TEST_123', True, True)
-                result = settings.get_setting('TEST_123')
+                settings.get_setting('TEST_123')
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -1,7 +1,10 @@
 
-from django.test import TestCase, override_settings
+import warnings
 
-from django_filters.conf import is_callable, settings
+from django.test import TestCase, override_settings
+from django.conf import settings as dj_settings
+
+from django_filters.conf import is_callable, settings, DEPRECATED_SETTINGS, DEFAULTS
 
 
 class DefaultSettingsTests(TestCase):
@@ -89,3 +92,14 @@ class IsCallableTests(TestCase):
         self.assertFalse(is_callable(Class))
         self.assertTrue(is_callable(c))
         self.assertTrue(is_callable(c.method))
+
+
+class SettingsObjectTestCase(TestCase):
+
+    def test_get_setting_deprecated(self):
+        with override_settings(FILTERS_TEST_123=True):
+            with self.assertWarns(DeprecationWarning):
+                DEPRECATED_SETTINGS.append('TEST_123')
+                DEFAULTS['TEST_123'] = True
+                settings.change_setting('FILTERS_TEST_123', True, True)
+                result = settings.get_setting('TEST_123')
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -1,3 +1,5 @@
+from unittest import mock
+
 from django.test import TestCase, override_settings
 
 from django_filters.conf import (
@@ -97,10 +99,11 @@ class IsCallableTests(TestCase):
 
 class SettingsObjectTestCase(TestCase):
 
+    @mock.patch('django_filters.conf.DEPRECATED_SETTINGS', ['TEST_123'])
+    @mock.patch.dict('django_filters.conf.DEFAULTS', {'TEST_123': True})
     def test_get_setting_deprecated(self):
         with override_settings(FILTERS_TEST_123=True):
             with self.assertWarns(DeprecationWarning):
-                DEPRECATED_SETTINGS.append('TEST_123')
-                DEFAULTS['TEST_123'] = True
                 settings.change_setting('FILTERS_TEST_123', True, True)
-                settings.get_setting('TEST_123')
+                test_setting = settings.get_setting('TEST_123')
+                self.assertTrue(test_setting)
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -1,10 +1,15 @@
 
 import warnings
 
-from django.test import TestCase, override_settings
 from django.conf import settings as dj_settings
+from django.test import TestCase, override_settings
 
-from django_filters.conf import is_callable, settings, DEPRECATED_SETTINGS, DEFAULTS
+from django_filters.conf import (
+    DEFAULTS,
+    DEPRECATED_SETTINGS,
+    is_callable,
+    settings
+)
 
 
 class DefaultSettingsTests(TestCase):
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -2,13 +2,7 @@ from unittest import mock
 
 from django.test import TestCase, override_settings
 
-from django_filters.conf import (
-    DEFAULTS,
-    DEPRECATED_SETTINGS,
-    is_callable,
-    settings
-)
-
+from django_filters.conf import is_callable, settings
 
 class DefaultSettingsTests(TestCase):
 
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -4,6 +4,7 @@ from django.test import TestCase, override_settings
 
 from django_filters.conf import is_callable, settings
 
+
 class DefaultSettingsTests(TestCase):
 
     def test_verbose_lookups(self):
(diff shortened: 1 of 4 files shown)


tests/test_conf.py
```diff
@@ -89,3 +93,14 @@ def method(self):
...
         self.assertTrue(is_callable(c.method))
+
+
+class SettingsObjectTestCase(TestCase):
+
+    def test_get_setting_deprecated(self):
+        with override_settings(FILTERS_TEST_123=True):
+            with self.assertWarns(DeprecationWarning):
+                DEPRECATED_SETTINGS.append('TEST_123')
+                DEFAULTS['TEST_123'] = True
```
hartwork: This change seems to outlast `test_get_setting_deprecated` runtime. Maybe `unittest.mock.patch` would help here.

tests/test_filters.py
```diff
@@ -388,6 +388,12 @@ def test_conjoined_true(self):
         f = MultipleChoiceFilter(conjoined=True)
         self.assertTrue(f.conjoined)
 
+    def test_is_noop_false(self):
+        f = MultipleChoiceFilter(required=False)
+        f.always_filter = False
+        qs = mock.Mock(spec='filter')
+        self.assertFalse(f.is_noop(qs, ['value']))
```
hartwork: I'm unsure why `is_noop` does not just get `None, None` as input here — the method body does not seem to use `qs, ['value']`. Am I missing something?

tests/test_widgets.py
```diff
@@ -185,6 +185,34 @@ class W(SuffixedMultiWidget):
         }, {}, 'price')
         self.assertEqual(result, ['1', 'lt'])
 
+    def test_value_omitted_from_data(self):
+        class A(SuffixedMultiWidget):
+            suffixes = ['_b']
+
+        a = A(widgets=[BooleanWidget])
```
hartwork: Judging from https://github.com/carltongibson/django-filter/blob/5d1f6a512c0c8f092897bf805a3b3990451202e8/django_filters/widgets.py#L108 I think we want an instance of `BooleanWidget` here, not a class: the widget's `value_omitted_from_data` is a regular method that wants `self`.

tests/test_widgets.py
```diff
@@ -185,6 +185,34 @@ class W(SuffixedMultiWidget):
...
+
+        result = a.value_omitted_from_data(['a', 'b', 'c'], './.', 'test')
```
hartwork: I'm not sure about why `['a', 'b', 'c'], './.'` was chosen for `data, files` here. Would simpler `None, []` or so do as well?

tests/test_widgets.py
```diff
@@ -185,6 +185,34 @@ class W(SuffixedMultiWidget):
         }, {}, 'price')
         self.assertEqual(result, ['1', 'lt'])
 
+    def test_value_omitted_from_data(self):
+        class A(SuffixedMultiWidget):
+            suffixes = ['_b']
```
hartwork: The `_` is unexpected to me here. Just wondering if others would be surprised as well.

tests/test_widgets.py
```diff
@@ -185,6 +185,34 @@ class W(SuffixedMultiWidget):
...
+    def test_replace_name(self):
+        class A(SuffixedMultiWidget):
+            suffixes = ['test']
+
+        a = A(widgets=[BooleanWidget])
+
+        output = '<div name="test123_0"></div>'
+        index = 0
+        q = a.replace_name(output, index)
+        self.assertEqual(q, output.replace('0', 'test'))
```
hartwork: I'd compare to explicit `'<div name="test123_test"></div>'` to simplify. Just an idea.

tests/test_widgets.py
```diff
@@ -185,6 +185,34 @@ class W(SuffixedMultiWidget):
...
+
+        result = a.value_omitted_from_data(['a', 'b', 'c'], './.', 'test')
+
+        self.assertIsNotNone(result)
+
+    def test_replace_name(self):
+        class A(SuffixedMultiWidget):
+            suffixes = ['test']
+
+        a = A(widgets=[BooleanWidget])
```
hartwork: The widget class is not used below and an instance seems needed if it was used. How about `None` or `mock.Mock()` for simplicity?

tests/test_widgets.py
```diff
@@ -185,6 +185,34 @@ class W(SuffixedMultiWidget):
...
+        output = '<div name="test123_0"></div>'
+        index = 0
+        q = a.replace_name(output, index)
+        self.assertEqual(q, output.replace('0', 'test'))
+
+    def test_decompress_value_None(self):
+        class A(SuffixedMultiWidget):
+            suffixes = ['']
+
+        a = A(widgets=[BooleanWidget])
```
hartwork: Same comment regarding `BooleanWidget` as above

tests/test_widgets.py
```diff
@@ -185,6 +185,34 @@ class W(SuffixedMultiWidget):
...
+            suffixes = ['test']
+
+        a = A(widgets=[BooleanWidget])
+
+        output = '<div name="test123_0"></div>'
+        index = 0
+        q = a.replace_name(output, index)
+        self.assertEqual(q, output.replace('0', 'test'))
+
+    def test_decompress_value_None(self):
```
hartwork: Is the capital `N` in the method name intended? I see lower `_none` a few times above.

tests/test_filters.py
rpkilby: `is_noop` does use the value

https://github.com/carltongibson/django-filter/blob/a5b4d0299ec04dcb103dc37291001fb268107c66/django_filters/filters.py#L226

I think the `qs` is provided because someone may want to override the behavior of `is_noop`, and having access to the `qs` is likely useful.

tests/test_widgets.py
rpkilby: The `MultiWidget` base instantiate the widgets if provided as classes. So, there should be no issue with accidentally passing a class instead. See:

https://github.com/django/django/blob/d232fd76a85870daf345fd8f8d617fe7802ae194/django/forms/widgets.py#L804

philsupertramp: @rpkilby thanks for clarifying this. My intention was to keep the tests within `SuffixedMultiWidgetTests` consistent; hence I decided to stay with passing the class, instead of passing an instance.

philsupertramp:
@hartwork thanks for your comments!
Could you elaborate your idea behind https://github.com/carltongibson/django-filter/pull/1130#discussion_r334255012 ? I may miss something but which object would you mock to achieve which behavior?

hartwork:
> @hartwork thanks for your comments!
> Could you elaborate your idea behind [#1130 (comment)](https://github.com/carltongibson/django-filter/pull/1130#discussion_r334255012) ? I may miss something but which object would you mock to achieve which behavior?

As @rpkilby pointed out, I overlooked use of `['value']`. For `qs`, `mock.Mock(spec='filter')` seems overly complex — what's the idea here? I was thinking that plain `None` or `mock.Mock()` would do. What do you think?

philsupertramp:
> > @hartwork thanks for your comments!
> > Could you elaborate your idea behind [#1130 (comment)](https://github.com/carltongibson/django-filter/pull/1130#discussion_r334255012) ? I may miss something but which object would you mock to achieve which behavior?
> 
> As @rpkilby pointed out, I overlooked use of `value` / `['value']`. For `qs`, `mock.Mock(spec='filter')` seems overly complex — what's the idea here? I was thinking that plain `None` or `mock.Mock()` would do. What do you think?

@hartwork sorry, I was referring to the comment regarding another comment
> This change seems to outlast `test_get_setting_deprecated` runtime. Maybe `unittest.mock.patch` would help here.

I guess the focus on the comment behind the link was a bit misleading

hartwork:
> > This change seems to outlast `test_get_setting_deprecated` runtime. Maybe `unittest.mock.patch` would help here.
> 
> I guess the focus on the comment behind the link was a bit misleading

Let me try to rephrase it: My impression is that lines…
```
                DEPRECATED_SETTINGS.append('TEST_123')
                DEFAULTS['TEST_123'] = True
```
have global affect that is not reverted, unlike when using a flavor of `mock.patch`. So right now these changes might affect tests that are run after and may even cause flaky tests somewhere in the future. That's my concern.

philsupertramp:
@hartwork thanks again for your help, I've updated the test regarding your comment.

tests/test_conf.py
```diff
@@ -1,7 +1,13 @@
+from unittest import mock
 
 from django.test import TestCase, override_settings
 
-from django_filters.conf import is_callable, settings
+from django_filters.conf import (
+    DEFAULTS,
+    DEPRECATED_SETTINGS,
```
hartwork: These make the linter complain now:
```
tests\test_conf.py:5:1: F401 'django_filters.conf.DEFAULTS' imported but unused
tests\test_conf.py:5:1: F401 'django_filters.conf.DEPRECATED_SETTINGS' imported but unused
```

philsupertramp: thanks, didn't see that.

PR #1176, opened 2020-02-26
Fix descending choices building for OrderingFilter

### LINKED ISSUE #1175: Bug in descending choices building for OrderingFilter
I got a problem assigning custom labels to reverse ordering options with `field_labels` parameter. I think [this method](https://github.com/carltongibson/django-filter/blob/master/django_filters/filters.py#L739-L750) is incorrect in its `descending = ...` part. It looks up for `-param` instead of `-field`, as stated in the [docs](https://django-filter.readthedocs.io/en/master/ref/filters.html#orderingfilter): «Keep in mind that the key is the field name, and not the exposed parameter name».

I have a fix for this issue, including an upgrade of the current test case with this issue in mind. I will provide it ASAP. If anything is wrong with my fix, please let me know. I'd like it to work for everybody, not just me.

jock-tanner:
Fixes  #1175

### DIFF
--- a/django_filters/filters.py
+++ b/django_filters/filters.py
@@ -737,12 +737,13 @@ class OrderingFilter(BaseCSVFilter, ChoiceFilter):
         ])
 
     def build_choices(self, fields, labels):
+        fields = {param: field for field, param in fields.items()}
         ascending = [
             (param, labels.get(field, _(pretty_name(param))))
-            for field, param in fields.items()
+            for param, field in fields.items()
         ]
         descending = [
-            ('-%s' % param, labels.get('-%s' % param, self.descending_fmt % label))
+            ('-%s' % param, labels.get('-%s' % fields[param], self.descending_fmt % label))
             for param, label in ascending
         ]
 
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -1508,17 +1508,19 @@ class OrderingFilterTests(TestCase):
 
     def test_field_labels_descending(self):
         f = OrderingFilter(
-            fields=['username'],
+            fields=(('a', 'c'), ('b', 'd')),
             field_labels={
-                'username': 'BLABLA',
-                '-username': 'XYZXYZ',
+                '-a': 'BLABLA',
+                '-b': 'XYZXYZ',
             }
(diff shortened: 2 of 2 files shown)


django_filters/filters.py
```diff
@@ -737,12 +737,13 @@ def normalize_fields(cls, fields):
...
     def build_choices(self, fields, labels):
+        fields = {param: field for field, param in fields.items()}
         ascending = [
             (param, labels.get(field, _(pretty_name(param))))
-            for field, param in fields.items()
+            for param, field in fields.items()
         ]
         descending = [
-            ('-%s' % param, labels.get('-%s' % param, self.descending_fmt % label))
+            ('-%s' % param, labels.get('-%s' % fields[param], self.descending_fmt % label))
```
rpkilby: I think the better approach here is to just call `pretty_name` again, instead of inverting the dict again. Something like...
```suggestion
            ('-%s' % param, labels.get('-%s' % field, self.descending_fmt % _(pretty_name(param))))
```

need to check if that's correct though..

django_filters/filters.py
```diff
@@ -737,12 +737,13 @@ def normalize_fields(cls, fields):
...
             for param, label in ascending
```
rpkilby: I think the better approach here is to just call `pretty_name` again, instead of inverting the dict again. Something like...
```suggestion
            ('-%s' % param, labels.get('-%s' % field, self.descending_fmt % _(pretty_name(param))))
            for field, param in fields.items()
```

need to check if that's correct though..

rpkilby: hm. It's not pretty, but I find the switch back and forth between field<>param confusing. 

That said, this might be superseded by #1118 or similar PR.

rpkilby:
Hi @jock-tanner. I tried pushing a few changes to your branch, but was unable to. Closing in favor of #1183. Thanks!

jock-tanner:
@rpkilby OK now I see my timing with this PR was no good considering the upcoming `OrderingFilter` refactoring in https://github.com/carltongibson/django-filter/pull/1118. Thank you for taking your time nevertheless.

PR #1195, opened 2020-04-01
Fix class name in code sample

malerichc:
Change FilterSet subclass name so that it matches the filterset_class later in the sample:

https://github.com/carltongibson/django-filter/blob/c98e52d80d8deb626fe5f4826cf2872e79a86459/docs/guide/rest_framework.txt#L144


rpkilby:
Thanks!

PR #1203, opened 2020-04-21
Rewrite filtering based on Q objects

hodossy:
This PR rewrites the filters to use Q objects instead of subsequent `filter()` calls. The goal is to provide a way to support arbitraty logic for combining filters in filtersets. The idea is that the `filter_queryset()` function would receive a list of **Q** objects, and by default it calls `filter` with each. If overridden, any custom logic could be supported.

Related to #1167. Groups could be implemented based on this new interface as well, but some discussion is needed there.

Problems to work on:
- Distinct calls are sometimes made before, sometimes after the filtering. What is the reason behind it?

@carltongibson What do you think of this PR? Does it worth finishing? Could you help me sort out the distinct calls?

Still left:
- [ ] Filterset filtering rewrite
- [ ] Doc update

### DIFF
--- a/django_filters/filters.py
+++ b/django_filters/filters.py
@@ -136,13 +136,30 @@ class Filter(object):
             self._field = self.field_class(label=self.label, **field_kwargs)
         return self._field
 
+    def get_filter_predicate(self, value):
+        lookup = '%s__%s' % (self.field_name, self.lookup_expr)
+        return {lookup: value}
+
+    def _create_q_object(self, value):
+        q = Q(**self.get_filter_predicate(value))
+        return ~q if self.exclude else q
+
+    def get_q_objects(self, value):
+        if value in EMPTY_VALUES:
+            return (Q(), )
+        return (self._create_q_object(value), )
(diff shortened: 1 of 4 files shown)


rpkilby:
Hi @hodossy. This approach has been discussed before, but it's basically a nonstarter. While Q objects do have benefits (e.g., can OR them together), they also prevent you from using a whole set of queryset features, like annotations and ordering. As an example, it would not be possible to implement `OrderingFilter`, since that filter specifically needs to call `.order_by()`. This is the same issue as you're running into with `.distinct()`, except it's complicated by the fact that you `.order_by()` also requires arguments.

hodossy:
Yes, I have noticed `OrderingFilter`, and I have a solution for that, of course not so elegant, and I was also able to handle `distinct()` calls as well. (I am just not sure why some filters apply distinct before they filter, and some after. I also run into a problem with `FilterMethod`, as I am affraid I cannot keep backward compatibility with it (or just in really ugly way).

PR #1215, opened 2020-05-06
Add 'required' check to CSV fields

### LINKED ISSUE #1213: Required BaseInFilters aren't required
I'm not sure if this also applies to forms or not, I only tested in DRF.

```python
class StringInFilter(filters.BaseInFilter, filters.CharFilter):
    pass

class ThingFilterSet(filters.FilterSet):
    foo = StringInFilter(field_name="foo", required=True)
    ...
```

If I just use a regular `CharFilter` it works, and will show an error when accessing the view. But using `BaseInFilter`, it doesn't seem to care if it's missing.

rpkilby:
Fixes #1213 and #869.

#854 should have fixed `required` handling for when CSV field got a _list_ of empty values, but not when an empty value itself was provided.

### DIFF
--- a/django_filters/fields.py
+++ b/django_filters/fields.py
@@ -184,6 +184,9 @@ class BaseCSVField(forms.Field):
         return type(str('CSV%s' % widget.__name__), bases, {})
 
     def clean(self, value):
+        if value in self.empty_values and self.required:
+            raise forms.ValidationError(self.error_messages['required'], code='required')
+
         if value is None:
             return None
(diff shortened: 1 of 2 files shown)


overfl0:
Is anything preventing this PR from being merged?
I just stumbled upon the issue that this PR is fixing and it took me quite a while to figure out the root cause. This could have easily been prevented by merging this PR.

carltongibson (approved):
Yep, makes sense. No block, just finitude of human powers. Let's have it. 
I'll roll a release to coincide with Django 3.2

PR #1227, opened 2020-05-20
Require tests to pass on `latest`.

carltongibson:
What happens if we do this?


carltongibson:
That looked good. Let's have it.

PR #1268, opened 2020-09-18
Allow easy overriding of 'Unknown' in BooleanWidget

Real-Gecko:
I've encountered situation where client wanted "Not selected" instead of "Unknown" so I had to copy paste entire class just to override the string. This little PR adds ability to easily override the text like this
```python
is_active = django_filters.BooleanFilter(widget=BooleanWidget(empty_label='Not selected'))
```

### DIFF
--- a/django_filters/widgets.py
+++ b/django_filters/widgets.py
@@ -156,8 +156,8 @@ class BooleanWidget(forms.Select):
     This can be used for AJAX queries that pass true/false from JavaScript's
     internal types through.
     """
-    def __init__(self, attrs=None):
-        choices = (('', _('Unknown')),
+    def __init__(self, attrs=None, empty_label=None):
+        choices = (('', empty_label if empty_label else _('Unknown')),
                    ('true', _('Yes')),
                    ('false', _('No')))
         super().__init__(attrs, choices)
--- a/django_filters/widgets.py
+++ b/django_filters/widgets.py
@@ -157,7 +157,7 @@ class BooleanWidget(forms.Select):
     internal types through.
     """
     def __init__(self, attrs=None, empty_label=None):
-        choices = (('', empty_label if empty_label else _('Unknown')),
+        choices = (('', empty_label or _('Unknown')),
                    ('true', _('Yes')),
                    ('false', _('No')))
         super().__init__(attrs, choices)


django_filters/widgets.py
```diff
@@ -156,8 +156,8 @@ class BooleanWidget(forms.Select):
     This can be used for AJAX queries that pass true/false from JavaScript's
     internal types through.
     """
-    def __init__(self, attrs=None):
-        choices = (('', _('Unknown')),
+    def __init__(self, attrs=None, empty_label=None):
+        choices = (('', empty_label if empty_label else _('Unknown')),
```
michael-k, line 160:
```suggestion
        choices = (('', empty_label or _('Unknown')),
```

Real-Gecko:
Anyone?

carltongibson (changes requested):
Thanks for the thought. This isn't worth the extra API. Just use a `Select` widget.

gatsinski:
I opened a similar PR (#1217)  almost a year ago allowing custom choices. Not only the empty label but also for true and false.

I get your point and I agree we can just use `Select` widget but is it really not worth the extra API? Yes and No are not the only ways to say that something is True or False. If we are forced to settle with these predefined labels it will make the existence of `BooleanWidget` obsolete because we won't be able to use it for a lot of cases where it actually makes more sense than using Select widget. So we can then say that the whole `BooleanWidget` may not be worth the extra API. 

Django-filter is a library and as such it has to be flexible.

Guilouf:
I really dont get it, for `ChoiceField` we have the `empty_label` kwarg and `EMPTY_CHOICE_LABEL` setting, here https://github.com/carltongibson/django-filter/blob/main/django_filters/fields.py#L292,
but for `BooleanField` it's apparently out of question, as we also can see here: https://github.com/carltongibson/django-filter/issues/871
I can understand that you want a clean API, but why allow such options on some fields and not on the others @carltongibson ?

For future me or other people that struggle, example on how it have to be done as for now:
```python
from django.forms import  Select
from django_filters import BooleanFilter

nullable_boolean_choices = (
    (True, "Yes"),
    (False, "No"),
    (None, "Whatever"),
)

class SomeFilter(FilterSet):
    has_source = BooleanFilter(
        field_name='source',
        lookup_expr='isnull',
        widget=Select(choices=nullable_boolean_choices)
    )
```
In that example i want to filter objects that have at least one relation to a `Source` object, or no relation, or don't filter anything.
For what i tried it can't be done with a `ChoiceFilter` because of `isnull` not accepting strings.

PR #1269, opened 2020-09-21
Run tests with GitHub Actions

michael-k:
If I've [understood correctly](https://github.com/microsoft/azure-pipelines-image-generation#former-repository-for-azure-pipelines-vm-images-for-microsoft-hosted-cicd), Azure Pipelines are deprecated in favor of GitHub Actions.

My actual goal was to run the tests against Python 3.9, but they are currently failing.  Therefore I've extracted the switch to GitHub Actions and will provide the rest in a separate PR.  (If you're interested in the failing test, see https://github.com/michael-k/django-filter/runs/1144983249?check_suite_focus=true)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,78 @@
+---
+name: Tests
+on: [push, pull_request]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Ensure latest setuptools
+        run: |
+          python -m pip install --upgrade pip setuptools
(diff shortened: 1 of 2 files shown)


michael-k:
There are two things this PR does not cover, but I think they should be up for discussion anyway:

- Remove the configuration for unused CI/CD tools (Travis, Azure Pipelines).

  Out of habit I've instantly looked at .travis.yml, but only noticed a while later that it's no longer in use and instead Azure Pipelines are running.  Maintaining multiple CI/CD tools adds just unnecessary overhead.

- The tests aren't running on a schedule.

  In the past tests on main branches of various projects started to fail over time, eg. due to new releases of (test/lint) dependencies.  The failures might point to actual problems that should be fixed.  And it's also bad for contributors, because they have to argue¹ that their contribution didn't introduce the failure.

  ¹Those failures are usually totally out of scope of the PR; first-time contributors might not have the knowledge to fix them; and they also raise the bar for contributions

carltongibson (approved):
Super. Let's have this. Thanks @michael-k. 

Happy if you want to remove both Azure and Travis configs once this is up and running and we're happy. 

Schedule. Not sure I followed 100%, but sure, yep, let's have it. 😄

PR #1283, opened 2020-10-07
Improve german translation

christianwgd:
Fixed some wrong translations to german language, improved some others


gatsinski:
I think that you also need the binary file generated by `compilemessages`. They are part of the repo.

christianwgd:
Sorry, sometimes the .mo files are not part of repo and are created during release build. But here it is.

gatsinski:
Yes, and I believe they should always be part of the build process. This will avoid issues like #1320.

gatsinski:
Yes, and I believe they should always be part of the build process. This will avoid issues like #1320 but this is another topic.

carltongibson (approved):
Hi @christianwgd — thanks for this! 👍

PR #1291, opened 2020-10-23
Allow `is_valid` to raise a validation error when invalid

RealOrangeOne:
In the context of a DRF view, this allows `is_valid` to short-circuit the view and return a response stating the filter errors.

Currently this is possible to achieve with:

```python
        if not filterset.is_valid():
            raise ValidationError(filterset.errors)
```

But would be nice if the library could do that for you.

### DIFF
--- a/django_filters/filterset.py
+++ b/django_filters/filterset.py
@@ -2,6 +2,7 @@ import copy
 from collections import OrderedDict
 
 from django import forms
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.db.models.constants import LOOKUP_SEP
 from django.db.models.fields.related import (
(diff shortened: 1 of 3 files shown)


carltongibson (commented):
Hi. This isn't a change we want to add — the behaviour here belongs in the view, or the filterset backend that's in the view's domain. This is not least because (as you've noticed here) the type of the exception raised will very much depend on the type of view you're serving. 

See the rest_framework backend for how to handle this: 

https://github.com/carltongibson/django-filter/blob/b2b90061d640a35791f88b1fb4ef69d7f1bc263b/django_filters/rest_framework/backends.py#L94-L95

PR #1303, opened 2020-12-12
Rename "es_ES" localization to "es" and add missing strings

nicst35:
The es_ES localization can (and should) be the base translation for almost all Spanish locales.
Also, this PR updates the "es" .po files to keep all source references in sync.


carltongibson (approved):
Ok, thanks,

PR #1320, opened 2021-01-21
Fix Bulgarian translations

gatsinski:
It looks like the binary file for the Bulgarian translations is a copy of the one for the Russian language. It took me a while to realize that the .mo files are part of the repo and even if the actual translations are fine, the generated binary file outputs Russian words instead of Bulgarian.


gatsinski:
Just rebased against the master branch to fix the jobs.

carltongibson (approved):
HI @gatsinski — thanks. 👍

PR #1346, opened 2021-03-10
Crispy forms without DRF

panosangelopoulos:
…he two situations


carltongibson (changes requested):
This isn't something I want to add to the package. You can do it in your own project. 

I'd probably assign the helper to the form after calling `filterset.form` rather than doing it in the filter set, but up to you.

PR #1358, opened 2021-04-06
Small correction to documentation

timnyborg:
Point out declaring filters in a field list has a small (and useful) effect

See: https://github.com/carltongibson/django-filter/issues/1350


carltongibson (approved):
Great. Thanks,

PR #1364, opened 2021-04-21
Support comma separated values in MultipleChoiceFilter

alanoe:
Related to GitHub issue
https://github.com/carltongibson/django-filter/issues/137


carltongibson (commented):
This isn't right. If you want to use CSV then you should use a suitable widget that spits the input before it's validated by the form. 

(This isn't a change we'll accept here but in general, all changes need tests.)

PR #1366, opened 2021-04-28
Updated url in setup.py

smithdc1:
Hey Carlton --  just a small change as part of the migration to `main`.

I noticed that RTD continues to refer to `master`, I think that may need updating in their admin? 🕵️ 

https://django-filter.readthedocs.io/en/master/index.html


carltongibson (approved):
Super. Thanks.

PR #1376, opened 2021-05-06
usage.txt: Add required arguments to DecimalField constructor

perlun:
I noted that these were missing because `System check identified some issues`. I guess it makes sense to make these examples be "reasonably complete" so they can easily be copy-pasted and tested by our users.


carltongibson (approved):
OK, yes that makes sense. Thanks!

PR #1382, opened 2021-05-27
Fix broken documentaion links

AndrewGuenther:
Since the primary branch has been renamed from master to main, many
documentation links are now dead. This change is a find replace for all
https://django-filter.readthedocs.io/en/master references to
https://django-filter.readthedocs.io/en/main


carltongibson (approved):
Super. Thanks.

PR #1391, opened 2021-06-26
extra_kwargs for FilterSet generated filters

gaetano-guerriero:
Support `extra_kwargs` attribute in `FilterSet.Meta`.

This allows to pass custom arguments for autogenerated filters.

This is similar to django rest framework `ModelSerializer.Meta.extra_kwargs`.


carltongibson (commented):
Hi. This isn't API I want to add. If you need extra arguments, declare the filters. (Experience with DRF just revealed this caused confusion, and if it didn't exist it wouldn't now be added.)

Thanks.

PR #1405, opened 2021-08-06
Updated documentation on the on_delete option

mtb-beta:
Thanks for the great package.
I found some outdated information in the documentation, so I fixed it.

on_delete is required for django 2.0 and later.


carltongibson (approved):
Yep great thanks

Issue #1440, opened 2021-09-18
About "Keyword-only Arguments:" section in the documentation


mayukorin:
For the following codes in 
https://django-filter.readthedocs.io/en/stable/ref/filters.html#method

```
class F(FilterSet):
    """Filter for Books by if books are published or not"""
    published = BooleanFilter(field_name='published_on', method='filter_published')

    def filter_published(self, queryset, name, value):
        # construct the full lookup expression.
        lookup = '__'.join([name, 'isnull'])
        return queryset.filter(**{lookup: False})

        # alternatively, you could opt to hardcode the lookup. e.g.,
        # return queryset.filter(published_on__isnull=False)

    class Meta:
        model = Book
        fields = ['published']


# Callables may also be defined out of the class scope.
def filter_not_empty(queryset, name, value):
    lookup = '__'.join([name, 'isnull'])
    return queryset.filter(**{lookup: False})

class F(FilterSet):
    """Filter for Books by if books are published or not"""
    published = BooleanFilter(field_name='published_on', method=filter_not_empty)

    class Meta:
        model = Book
        fields = ['published']
```
I think it would be better to change it as following. 

1. If "published" field is present in "Book" model, then 
```
# published = BooleanFilter(field_name='published_on', method='filter_published')
published = BooleanFilter(field_name='published', method='filter_published')
```
2. If "published_on" field is present in "Book" model, then 
```
class Meta:
        model = Book
        # fields = ['published']
        fields = ['published_on']
```
I've made pull requests #, # for each pattern.

### Reason
I think field_name in BooleanFilter should match the field name in the model.

### Test

I used the code in 
https://django-filter.readthedocs.io/en/stable/ref/filters.html#method
as a reference and tested it.
```
# models.py
from django.db import models


class Author(models.Model):
    name = models.CharField(max_length=255)


class Book(models.Model):
    name = models.CharField(max_length=255)
    published = models.DateTimeField(null=True)
```
```
# serializers.py
from rest_framework import serializers
from .models import Book


class BookSerializer(serializers.ModelSerializer):
    class Meta:
        model = Book
        fields = "__all__"

```
```
# views.py
from django_filters import BooleanFilter
from django_filters import rest_framework as filters
from rest_framework import generics
from .serializers import BookSerializer
from .models import Book


class F(filters.FilterSet):

    published = BooleanFilter(field_name="published_on", method="filter_published")

    def filter_published(self, queryset, name, avlue):
[...]

PR #1441, opened 2021-09-18
fix field_name of BooleanFilter

mayukorin:
First, please see the Issue #1440.
This is about #1440  "1. published field exists in Book model".
I think it would be better to change it as following. 
```
# published = BooleanFilter(field_name='published_on', method='filter_published')
published = BooleanFilter(field_name='published', method='filter_published')
```
Modified code was tested with following change on the test section of #.

```
# views.py
from django_filters import BooleanFilter
from django_filters import rest_framework as filters
from rest_framework import generics
from .serializers import BookSerializer
from .models import Book


class F(filters.FilterSet):

    # published = BooleanFilter(field_name="published_on", method="filter_published")
    published = BooleanFilter(field_name="published", method="filter_published")
   ....

```
The result was successful.
```
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
published_on__isnull
.
Ran 1 test in 0.469s

OK
Destroying test database for alias 'default'...
```

### DIFF
--- a/docs/ref/filters.txt
+++ b/docs/ref/filters.txt
@@ -66,7 +66,7 @@ transformation and empty value checking should be unnecessary.
 
     class F(FilterSet):
         """Filter for Books by if books are published or not"""
-        published = BooleanFilter(field_name='published_on', method='filter_published')
+        published = BooleanFilter(field_name='published', method='filter_published')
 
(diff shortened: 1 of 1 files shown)


carltongibson (commented):
I don't think this change is necessary. Thanks.

PR #1445, opened 2021-09-24
setup.cfg: Use underscore in license_file key

mgorny:
Use underscore in the setup.cfg key to fix the following warning:

> Usage of dash-separated 'license-file' will not be supported in future versions. Please use the underscore name 'license_file' instead


carltongibson (approved):
Yep. Great. Thanks. 👍

PR #1451, opened 2021-10-16
Remove assertions/warnings about code deprecated in version 2.1

abbottc:


abbottc:
Incidentally, due to this PR, the class `RenameAttributesBase` in `django_filters/utils.py` is no longer used anywhere. I didn't remove that class, assuming it may be something you want to keep for possible future use, but if you want it removed I can do that as well.

carltongibson (approved):
Nice. Thanks @abbottc

mochounan:
https://github.com/carltongibson/django-filter/commit/3dfed5d6755f6e5e25e967e371eb61774eeed015#r77608630
at present, the pypi official website has appeared 22.1. This problem leads to the query not accessing the orm

Is there something wrong with my use?

PR #1462, opened 2021-12-13
Improved DateRangeFilter by tackling timezone difference.

xemexpress:
It will improve this [issue](https://github.com/carltongibson/django-filter/issues/1461) maybe?

Thank you!


carltongibson (commented):
Can you add a test case (or some test cases) showing the incorrect behaviour that is addressed by the change please?

carltongibson (commented):
As per the ticket, this goes against the advice in the [Django Time Zone docs](https://docs.djangoproject.com/en/4.0/topics/i18n/timezones/) to do all calculations in UTC.

PR #1539, opened 2022-10-26
Undefined filter exception msg add model

ibuler:
This makes it easier for us to troubleshoot


carltongibson (commented):
The point of the message is that field is not on the model, so the "of ..." doesn't make sense. 

The context should be clear enough from the traceback I think.  Thanks.

PR #1540, opened 2022-10-27
fix(QueryArrayWidget): value_from_datadict should not mutate input data

### LINKED ISSUE #1047: QueryArrayWidget mutates passed data
I think I've faced a small issue with the `QueryArrayWidget`, especially with this line: https://github.com/carltongibson/django-filter/blob/master/django_filters/widgets.py#L242, as described below:

```python
class QueryArrayWidget(BaseCSVWidget, forms.TextInput):

    def value_from_datadict(self, data, files, name):
        if not isinstance(data, MultiValueDict):
            for key, value in data.items():
                # treat value as csv string: ?foo=1,2
                if isinstance(value, str):
                    data[key] = [x.strip() for x in value.rstrip(',').split(',') if x]
```

The last line of this snippet will mutate the initial QueryDict, which can have unforeseen consequences. An easy fix could be to simply create a new dict and I can provide a PR with that if we agree on a fix :)

Edit: temporary workaround in my project is to subclass like that:

```python
class QueryArrayWidget(widgets.QueryArrayWidget):
    """
    Until https://github.com/carltongibson/django-filter/issues/1047 is fixed
    """

    def value_from_datadict(self, data, files, name):
        data = data.copy()
        return super().value_from_datadict(data, files, name)
```

legau:
Resolves #1047

### DIFF
--- a/django_filters/widgets.py
+++ b/django_filters/widgets.py
@@ -246,6 +246,7 @@ class QueryArrayWidget(BaseCSVWidget, forms.TextInput):
 
     def value_from_datadict(self, data, files, name):
         if not isinstance(data, MultiValueDict):
+            data = data.copy()
             for key, value in data.items():
                 # treat value as csv string: ?foo=1,2
                 if isinstance(value, str):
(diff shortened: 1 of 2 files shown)


carltongibson (approved):
Great. Thanks.

PR #1564, opened 2023-02-23
Remove hardcode of rest_framework crispy form template_pack

### LINKED ISSUE #1537: Using crispy-bootstrap5 with Django-filter and DRF
FilterSet in django_filters.rest_framework.filterset.py defines the template pack to use. It looks like bootstrap3 is hardcoded here:

```python3
        if compat.is_crispy():
            from crispy_forms.helper import FormHelper
            from crispy_forms.layout import Layout, Submit

            layout_components = list(form.fields.keys()) + [
                Submit('', _('Submit'), css_class='btn-default'),
            ]
            helper = FormHelper()
            helper.form_method = 'GET'
            helper.template_pack = 'bootstrap3'
            helper.layout = Layout(*layout_components)
```

This breaks `crispy_bootstrap5`. Would it not be nicer to do?:
```python
from django.conf import settings
```
```python3
        if hasattr(settings, 'CRISPY_TEMPLATE_PACK'):
            helper.template_pack = settings.CRISPY_TEMPLATE_PACK
        else:
            helper.template_pack = 'bootstrap3'
```

Or is there an other place to set the template pack?

bckohan:
Fixes #1537.

The template_pack is hardcoded to 'bootstrap3' in rest_framework's FilterSet. This change simply removes the setter allowing crispy forms to determine what the template pack should be based on the crispy forms configuration or whichever the default pack is for the installed version of crispy forms.

The Layout objects used are generic and should be compatible with any template pack, but as before if users want to customize the crispy form they can override the form property.

This change is necessary because some template packs step on each other. In this case bootstrap3 and bootstrap5 are incompatible. Simply installing the bootstrap5 template pack is enough to crash browsable DRF rendering.

### DIFF
--- a/django_filters/rest_framework/filterset.py
+++ b/django_filters/rest_framework/filterset.py
@@ -34,7 +34,6 @@ class FilterSet(filterset.FilterSet):
             ]
             helper = FormHelper()
             helper.form_method = "GET"
-            helper.template_pack = "bootstrap3"
             helper.layout = Layout(*layout_components)
 
             form.helper = helper


carltongibson (approved):
OK, thanks. 

I'm prepping the next release to coincide with Django 4.2 so this will be part of that.

PR #1587, opened 2023-06-27
Update widgets.py

zhaojiejoe:
make sure that the ret in the correct order


carltongibson (commented):
Can you add a test case showing what this is meant to fix?

zhaojiejoe:
I made a new rp with the test case.

PR #1588, opened 2023-06-27
Update widgets.py

zhaojiejoe:


zhaojiejoe:
sometimes we need the correct order in the query array. Set will change the order.
>>> ret=[2022, 2021]
>>> list(set(ret))
[2021, 2022]
>>>

carltongibson (commented):
This duplicates the existing tests, which already pass. It doesn't look necessary.

PR #1611, opened 2023-10-10
Example snippet fix: explicit check against None

m2-farzan:
The previous script ignores any initial value that has a falsy value while the e pected behavior is to only skip them if they are not present in filter args (i.e. None value)


carltongibson (commented):
I don't really want to change a several year old example from pre-2.0. Thanks.

PR #1623, opened 2023-11-14
Fix DeprecationWarning for pkgutil.find_loader

wmorrell:
The `find_loader` function is deprecated in Python 3.12 and slated for removal in Python 3.14. The function is a wrapper around `importlib.util.find_spec`, and returning the `loader` attribute from that call. The code isn't using the loader, and only checking that the import path exists, so we may use the wrapped function directly, instead.

fixes: #1616


carltongibson (approved):
Great. Thanks. 🎁

PR #1642, opened 2024-03-08
fix: Remove usage of deprecated is_iterable util

adamantike:
Django announced [1] that the `django.utils.itercompat` module is deprecated since Django 5.1.

Instead, the same behavior can be achieved with an instance comparison against `Iterable` [2].

[1] https://docs.djangoproject.com/en/dev/releases/5.1/#features-deprecated-in-5-1
[2] https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable

### DIFF
--- a/django_filters/filters.py
+++ b/django_filters/filters.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from collections.abc import Iterable
 from datetime import timedelta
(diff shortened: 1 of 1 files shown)


carltongibson (approved):
Yes, looks good.

PR #1651, opened 2024-03-26
remove `...` in rest_framework.txt to make life easy

vainl:
Someone silly like me, who just copies and posts the code snippet with the ..., will get an exception `AttributeError: 'ellipsis' object has no attribute 'rsplit'`. So let's just remove it and make life easy.

### DIFF
--- a/docs/guide/rest_framework.txt
+++ b/docs/guide/rest_framework.txt
@@ -39,7 +39,6 @@ If you want to use the django-filter backend by default, add it to the ``DEFAULT
 
     # settings.py
     INSTALLED_APPS = [
-        ...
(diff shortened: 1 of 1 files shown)


carltongibson (commented):
It's meant to imply there's more there. I'd be happy if you replaced it with a commment `# …`

jeremystretch:
IMO it's very important to retain some indicator that the example does not represent the complete configuration, to avoid confusion.

vainl:
> It's meant to imply there's more there. I'd be happy if you replaced it with a commment `# …`

Done. I just a comment.

carltongibson (approved):

PR #1660, opened 2024-05-14
Remove timezone on DateRangeField

Thutmose3:
https://github.com/carltongibson/django-filter/issues/1659


carltongibson (commented):
Can you add regression tests for the change here please?  Thanks.

PR #1730, opened 2025-10-05
Updated testing for Django 6.0.

carltongibson:


tox.ini
```diff
@@ -19,10 +17,8 @@ commands = coverage run --parallel-mode --source django_filters ./runtests.py --
 setenv =
        PYTHONDONTWRITEBYTECODE=1
 deps =
-    django42: Django>=4.2,<5.0
-    django50: Django>=5.0,<5.1
-    django51: Django>=5.1,<5.2
-    django52: Django>=5.2a1,<6.0
+    django52: Django>=5.2,<6.0
+    django60: Django>=6.0a1,<7.0
```
michael-k, line 21: Shouldn't this have been 6.1 instead of 7.0?

```suggestion
    django60: Django>=6.0a1,<6.1
```

pyproject.toml
```diff
@@ -16,19 +16,17 @@ classifiers = [
...
-    "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
 ]
-requires-python = ">=3.9"
-dependencies = ["Django>=4.2"]
+requires-python = ">=3.10"
+dependencies = ["Django>=5.2"]
```
christianklemp, line 29: Why would support for an LTS version of Django be dropped before the LTS has reached end-of-life? Seems like it should still be supported based on what's in the README?

> Support for Python and Django versions will be dropped when they reach end-of-life. Support for Python versions will be dropped when they reach end-of-life, even when still supported by a current version of Django.

carltongibson: This is in preparation for Django 6.0. As per [the Django release notes](https://docs.djangoproject.com/en/dev/releases/6.0/#third-party-library-support-for-older-versions-of-django): 

> Following the release of Django 6.0, we suggest that third-party app authors drop support for all versions of Django prior to 5.2. 

There are no significant code changes. Existing versions will continue to work with Django 4.2

christianklemp: That's interesting since 4.2 is LTS. I'm unfortunately restricted to the 4.2 LTS version currently until early next year. Once Renovate bumped to the new version of `django-filter` with this change, I started getting an unresolvable conflict error when trying to pip install due to Django being pinned at 4.2 still. I'll have to prevent Renovate from updating this package until I can upgrade early next year.

Thank you for the clarification and all your work on this package!

carltongibson: Yes, just pin django-filter to 25.1 until you're ready to update Django.

suricactus:
Hey, @carltongibson . Is there a specific technical reason why we already need to drop support for Django 4.2 which is still maintained until April 2026 (https://www.djangoproject.com/download/) ?

If there is a extra maintanence burden, I would be interested to hear if our team can support you until Django 4.2 is officially sunsetting.

carltongibson:
> This is in preparation for Django 6.0. As per [the Django release notes](https://docs.djangoproject.com/en/dev/releases/6.0/#third-party-library-support-for-older-versions-of-django): 
> 
> 
> 
> > Following the release of Django 6.0, we suggest that third-party app authors drop support for all versions of Django prior to 5.2. 
> 
> 
> 
> There are no significant code changes. Existing versions will continue to work with Django 4.2

suricactus:
@carltongibson yes, read that, but unfortunately the dependencies are defined in different way:

https://github.com/carltongibson/django-filter/blob/main/pyproject.toml#L29

```
[/tmp/django-filter]$ uv venv
Using CPython 3.12.3 interpreter at: /usr/bin/python3
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate
[/tmp/django-filter]$ uv pip install django==4.2.25 django-filter==25.2 
  × No solution found when resolving dependencies:
  ╰─▶ Because django-filter==25.2 depends on django>=5.2 and you require django==4.2.25, we can conclude that your requirements and django-filter==25.2 are incompatible.
      And because you require django-filter==25.2, we can conclude that your requirements are unsatisfiable.
```

Reproducible with pure `pip` too.

carltongibson:
Yes. Pin to v25.1 until you're ready to update Django. 

The 5.2 LTS is already available to you.

PR #1744, opened 2025-12-23
Fix #1734: Handle null_value correctly in BooleanFilter

### LINKED ISSUE #1734: filters.py line 191 None should be True

Olsion23:
This PR fixes issue #1734 where BooleanFilter raised TypeError when null_value was used.
Now None/null_value is correctly treated as True, and the filter works without errors.

- Updated ChoiceFilter to pop null_value from kwargs
- Added test for None/null_value handling


carltongibson:
> 6236 files changed. 

Can you fix up the PR, so I can look at it sensibly please?

PR #1750, opened 2026-05-09
Fill missing translations across 8 locales

efe:
I was scanning django-filter's locale files for missing translations and found several empty `msgstr` entries.

The changes cover Argentinian Spanish (`es_AR`), Belarusian (`be`), Bulgarian (`bg`), Danish (`da`), German (`de`), Greek (`el`), Italian (`it`), and Russian (`ru`).

One thing worth flagging: the Greek locale file had `Language: de` set incorrectly in the header — corrected to `Language: el`. Also, `is null` in Russian was left as `is null` since the term is commonly used as-is by Russian developers, but happy to change this to `является нулевым` or similar if you prefer a fully localised string.

### How these were generated

I built a platform called [Yesglot](https://yesglot.com) that automates translation of Django `.po` files. The output was reviewed before submitting — the Greek header fix was caught during that review.

### What's next

If this PR is welcome, I'd like to follow up with new locale files for languages that are completely missing from the repo: Turkish (`tr`), Dutch (`nl`), Polish (`pl`), Arabic (`ar`), Japanese (`ja`), Korean (`ko`), Traditional Chinese (`zh_TW`), Spanish Spain (`es`), Hungarian (`hu`), and Swedish (`sv`).

Happy to adjust any strings or split this into per-language PRs if that's easier to review.


carltongibson:
Hi @efe. I'm wondering how to review these suggestions. (The Greek header change looks correct, perhaps we can have that as an immediate quick fix separately?)

Feeling a dearth of native speakers...

efe:
Hello @carltongibson! Totally understand the review concern.

I'm a native Turkish speaker so I can personally vouch for `tr` if we add it. For the others, happy to take whatever approach you're most comfortable with — merging with a note inviting corrections, opening per-language issues for community review, or simply waiting. Happy to follow your lead.

And yes, happy to open the Greek header fix as a separate PR if that's easier! https://github.com/carltongibson/django-filter/pull/1751

Have a good day

carltongibson:
Hey @efe, so just a high-level question: in general are technical terms like null better translated or not? (I'm not sure the right answer is there, but it seems these entries were left deliberately given that the other terms are translated?). 

(I'm trying to work out the best approach here.)

efe:
Hello @carltongibson, That's a though question but I've thought about it.

In general, I think showing `null` in a UI is not ideal regardless of language, unless it's a purely engineer-facing product. Terms like "empty", "not set", "not available", or "unknown" tend to be more human-readable and less ambiguous for end users. However, `django-filter` is one of the exceptions. As `django-filter` user, I would expect "null mı?" (semi-Turkish) version. ("mı?" means "is?")

If keeping `null` as is is preferred way, I'd kindly suggest explicitly setting `msgstr` to `null` in each locale rather than leaving it empty. An empty string might be ambiguous: it's hard to tell if it was intentionally skipped or just forgotten.

Happy to go either way.

carltongibson:
@efe ok, shall we go for that suggestion then? (Keeping null but filling empty strings there.) I happy if you want to update this PR for that. Thanks!

PR #1758, opened 2026-08-16
fix: preserve submitted order in QueryArrayWidget.value_from_datadict

### LINKED ISSUE #1757: Multi-field ordering silently sorts by the wrong field when using QueryArrayWidget (value order not preserved)
## The issue

If you use `QueryArrayWidget` as the widget for a multi-value filter where **order matters** — most notably `OrderingFilter`, which supports multi-field sorting via `qs.order_by(*ordering)` — the order you submit values in isn't preserved. `?ordering[]=price&ordering[]=-name` (sort by price, then by name descending as a tiebreaker) can silently come back sorted as `-name,price` instead, with no error or warning — it just quietly sorts by the wrong field first.

## Root cause

`value_from_datadict` (`django_filters/widgets.py`) ends with:

```python
return list(set(ret))
```

Converting to a `set` before returning discards the order values were submitted in — Python sets aren't ordered by insertion:

```python
>>> list(set(['3', '1', '2', 'name', '-price']))
['-price', '2', '3', '1', 'name']
```

This is surprising for a widget documented as implementing "request query array notation" (`?foo[]=bar&foo[]=baz`) — array notation implies order is significant, and the docstring's own examples (`?foo=1&foo=2 -> foo = [1, 2]`) show an ordered result the actual return value doesn't honor.

## Reproduced

Environment: Python 3.13.3, Django 6.1, django-filter 26.1 (current `main`, commit `5128b8f`).

```python
import django_filters
from django_filters.widgets import QueryArrayWidget

    ordering = django_filters.OrderingFilter(
        fields=("price", "name"),
        widget=QueryArrayWidget,
    )
    class Meta:
        model = Product
        fields = []

# Client requests: sort by price ascending, then name descending as a tiebreaker
f = ProductFilter(data={"ordering[]": ["price", "-name"]}, queryset=Product.objects.all())
print(f.is_valid())
print(str(f.qs.query).split("ORDER BY")[-1])
```

Output on unpatched `main`:

```
True
 "tests_product"."name" DESC, "tests_product"."price" ASC
```

Requested `price, -name`; the generated `ORDER BY` came back as `name DESC, price ASC` — completely reversed. No error, no warning — the request just gets silently re-sorted by the wrong field first.

## Expected behavior

The returned list should preserve the order values were submitted in, so the `ORDER BY` above should read `"tests_product"."price" ASC, "tests_product"."name" DESC`. Deduplication is fine (and already documented) — just not the reordering that comes along with it.

## Suggested fix

Deduplicate while preserving order, e.g. `list(dict.fromkeys(ret))` instead of `list(set(ret))`.

## Related

Distinct from #1685 / #1686 (about the `MultiValueDict`/`QueryDict` detection branch earlier in the same method) — that fix doesn't touch the final `set()` call, so this would still reproduce afterward. Happy to send a PR for this one.

juneja-varun:
Fixes #1757.

`QueryArrayWidget.value_from_datadict` deduplicates parsed values by converting to a `set`, which doesn't preserve the order they were submitted in. That's surprising for a widget documented as implementing 'query array notation' where order matters, and it breaks `OrderingFilter` when it's backed by this widget - `?ordering=price,-name` can silently end up sorting by the wrong field first.

Fixed by using `dict.fromkeys()` instead of `set()` to dedupe while preserving order of first occurrence. Added a regression test confirming it fails with the scrambled order on unpatched code and passes with the fix.


juneja-varun:
Just checking in on this one — happy to make any changes if useful.

carltongibson:
@juneja-varun I havent had time to review yet. I've been flooded by a wave of AI generated PRs.

juneja-varun:
No worries at all — totally understand, take whatever time you need. Happy to answer any questions on this one whenever you get to it.

PR #1759, opened 2026-08-17
Fix CSV 'in'/'range' filters on multi-value fields (ManyToMany, MultipleChoiceFilter)

### LINKED ISSUE #1084: CSV filters incompatible with multiple choice fields
See https://github.com/carltongibson/django-filter/issues/976#issuecomment-422585413.

juneja-varun:
Fixes #1084.

I hit this building a filter for a many-to-many field: `GET /projects/?contacts__in=1,2` is the documented way to expose an `__in` lookup via `Meta.fields`, but it's completely broken for M2M/multiple-choice fields - raises `ValidationError: Enter a list of values.`, or once that's worked around, `TypeError: '...' object is not iterable`.

Two separate bugs: `BaseCSVField.clean()` cleaned each CSV value independently, which doesn't work for fields like `MultipleChoiceField` that expect the whole list in one `clean()` call. And the dynamically-generated filter class for this case resolves `.filter()` via MRO to `MultipleChoiceFilter.filter()` (per-value OR-combining) instead of the base list/range semantics `BaseCSVFilter` actually needs.

Fixed both, added regression tests confirming each fails on unpatched code and passes with the fix, full suite green.


juneja-varun:
Checking in on this one too — happy to make any changes if useful.

carltongibson:
@juneja-varun I havent had time to review yet. I've been flooded by a wave of AI generated PRs.

juneja-varun:
No rush at all — happy to answer any questions on this one whenever you get a chance to look.

